PR #5, opened 2025-02-01
Swedish translation language file

Ladida1:
Translation into Swedish.

All strings are translated. Might need some refinement after testing it longer, in that case I will submit a followup PR.


andreadegiovine:
Please update your translation for version 0.2.12

PR #60, opened 2025-03-07
Update es.json and fix the error in config_flow.py

xplore93:


xplore93:
Each time it is more and more difficult to update the translations. My proposal is to add new/modify keys in all JSON with the value in english. That way, the integration will work without errors and the translation will be made faster.

Right now, in spanish language the integration doesn't work because of this:

```
File "/config/custom_components/stellantis_vehicles/config_flow.py", line 96, in async_step_oauth
    oauth_label = self.get_translation("component.stellantis_vehicles.config.step.oauth.data.oauth_code").replace(" ", "_").upper()
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'replace'
```

andreadegiovine:
> Each time, it becomes increasingly difficult to update the translations. My suggestion is to add new or modify existing keys in all JSON files with the English value. This way, the integration will work without errors, and translations can be made more quickly.
> 
> Currently, the integration doesn't work due to the following reason:
> 
> ```
> File "/config/custom_components/stellantis_vehicles/config_flow.py", line 96, in async_step_oauth
>     oauth_label = self.get_translation("component.stellantis_vehicles.config.step.oauth.data.oauth_code").replace(" ", "_").upper()
>                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> AttributeError: 'NoneType' object has no attribute 'replace'
> ```

i think to add default lang (en) value

PR #88, opened 2025-03-18
Multiple vehicle token refresh

MoellerDi:
~~It seems that sometimes the token expires due to network delay or processing time, resulting in an "Invalid_Grant" error.
This PR gives a little more time and should refresh the token before it expires.~~

If multiple vehicles are configured in the mobile app, the integration might sometimes tries to refresh the same access_token twice causing an "Invalid_Grant" error and triggers "needs reauthorize/reconfigure" in HA.
This PR is fixing it by preventing simultaneous read/update the access/refresh-tokens. However (IMHO) a bit more complex refactoring might be needed to support multiple vehicles and to enable a better solution later.

### DIFF
--- a/custom_components/stellantis_vehicles/stellantis.py
+++ b/custom_components/stellantis_vehicles/stellantis.py
@@ -222,6 +222,8 @@ class StellantisVehicles(StellantisBase):
         self._callback_id = None
         self._mqtt = None
         self._mqtt_last_request = None
+        self._lock_refresh_token = asyncio.Lock()
+        self._lock_refresh_mqtt_token = asyncio.Lock()
 
     def set_entry(self, entry):
         self._entry = entry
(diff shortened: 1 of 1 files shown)


andreadegiovine:
Currently i use the refresh interval to refresh tokens befor expiration

https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/blob/6316dcbacb4ce1de1cbf6b8b96f1bb5cfa763061/custom_components/stellantis_vehicles/stellantis.py#L285

MoellerDi:
yes, but IMHO it needs some buffer time to allow time drift or network/processing delay. I'm almost sure, the "invalid_grant" error is cause by expired token.

andreadegiovine:
@MoellerDi have you any evidence? Please share here any logs of error (include the row datetime) and the token expiration

MoellerDi:
@andreadegiovine Not yet. I'll add debugging code to my system, gather the requested evidence, and update this PR accordingly.

MoellerDi:
@andreadegiovine can you please let me know if this PR is going in the right direction or if you would like a different implementation?

andreadegiovine:
@MoellerDi unfortunally i cant test multiple vehicles account (i have only one), please make changes on your fork and use it on your hass instance.

When i have time i do a refactoring for manage multiple vehicles in the best way.

The goal is to have one manager (StellantisVehicles) for multiple vehicles to improve the token expiration logic, at the moment the last vehicle loads by this integration was setted on manager (StellantisVehicles) and this create a bug over token refreshing.

Leave this pr open as i can work on this, if you make the refactoring and it work fine add here a comment and open new pr

MoellerDi:
I'll call it a day! At least the changes are now running on my Hass instance, but I'm still trying to familiarize myself with your code :-) I hope to get a bit more done tomorrow.

MoellerDi:
@andreadegiovine do you think it's a good idea to merge this PR as is to prevent the "Invalid_Grant" error and the triggering of "reauthorization/reconfiguration" warning in HA? In my opinion, it would be better to perform any refactoring in a new PR anyway.

MoellerDi:
This PR was replaced by #117

PR #109, opened 2025-03-24
Just a small fix in StellantisVehicleCoordinator

MoellerDi:
IMHO it's better to use self._vehicle["type"] :-)


andreadegiovine:
> IMHO it's better to use self._vehicle["type"] :-)

its the same data, the metod get_config read from a property that include vehicle data (see set_vehicle method), you preffer this?

MoellerDi:
This PR was replaced by #117

PR #110, opened 2025-03-24
Multiple vehicle support

MoellerDi:
This PR enable multiple vehicle support and is based on in PR #108

PR #108, opened 2025-03-24
Refactoring token refresh

MoellerDi:
This PR is refactoring the token refresh to ensure a valid token is available whenever make_http_request is called.

PR #117, opened 2025-03-28
Add multiple vehicle support and refactoring token refresh

MoellerDi:
This PR is replacing: #88, #108, #109 and #110

This PR is implementing multiple vehicle support as well as an (IMHO) improved token refresh handling.
Once this PR get merged, I like to further optimize the mqtt token refresh handling as it seems it's only needed for establishing the mqtt connection and doesn't need a refresh for the commands. Also I plan to review the general update_interval handling to see if it could be strictly separated for multiple vehicle.

Anyway, this PR was tested on my HA system for several days. All sensors, commands works as before (well now for both of my cars) and no errors whatsoever so far.

The following have been tested:
- [x] New installation – tested and working
- [x] Sensors – tested and working for two cars
- [x] Commands – tested and working for two cars
- [x] 1 week without an error log – no errors for >7 days


MoellerDi:
...another catch. I've met the "1 week error-free" requirement. Please let me know if you need anything else.
Cheers,
Dirk

koomik:
Hi Dirk, do you mean that you haven't got any manual reauth process ?

MoellerDi:
> Hi Dirk, do you mean that you haven't got any manual reauth process ?

@koomik yes, no reauth process required, no error. it just works :-)
This PR got merged and 0.4.0 is released - have fun

custom_components/stellantis_vehicles/stellantis.py
```diff
@@ -487,7 +486,8 @@ def _on_mqtt_message(self, client, userdata, msg):
 
     async def send_mqtt_message(self, service, message, vehicle, store=True):
         _LOGGER.debug("---------- START send_mqtt_message")
-        await self.refresh_mqtt_token(force=(store == False))
+        #TODO: check/confirm if a mqtt refresh_token is really needed here
+        await self.refresh_tokens(force=(store == False))
```
MoellerDi, line 490: @andreadegiovine your commit replaced refresh_mqtt_token() with refresh_tokens() .....
That's why commands are not working in 0.4.0 :-(

custom_components/stellantis_vehicles/stellantis.py
```diff
@@ -110,18 +112,16 @@ def apply_headers_params(self, headers):
...
         query_params = []
         for key in params:
             value = params[key]
             query_params.append(f"{key}={value}")
         query_params = '&'.join(query_params)
-        return self.replace_placeholders(f"{url}?{query_params}")
+        return self.replace_placeholders(f"{url}?{query_params}", vehicle)
 
     async def make_http_request(self, url, method = 'GET', headers = None, params = None, json = None, data = None):
-        _LOGGER.debug("---------- START make_http_request")
```
andreadegiovine, line 122: There are a specific reason to remove start/end http request log?

MoellerDi: Just to avoid double logging as it's re-added in line 286/290

custom_components/stellantis_vehicles/stellantis.py
```diff
@@ -383,27 +384,26 @@ async def get_vehicle_last_trip(self, page_token=False):
 #         _LOGGER.debug("---------- END get_vehicle_trips")
 #         return vehicle_trips_request
 
-    async def refresh_tokens(self, force=False):
```
andreadegiovine, line 386: Initially the refresh token methods was separated, but sometimes the mqtt token try to refresh after expiration because the message frequency skip expiration, therefore i will use a wrap method to refresh both tokens on any event (the most frequently is vehicle status http request), what do you think if we restore the wrap method and verify the mqtt token as frequently as the http token?

andreadegiovine:
Great job, I would have tried a different solution to manage multiple vehicles, in my opinion the problem was due to the saving of the "config" property which, being a deep copy, was not updated in the other instances of the class, so I would have added a level before the vehicle coordinator that managed the saving of the config.

Thank you, you will be included among the contributors on the main page of the repo.

custom_components/stellantis_vehicles/stellantis.py
MoellerDi: IMHO there is no need to refresh both tokens on any event. The refresh-token can be used at any time to get a fresh and valid access-token whenever needed. No need to maintain a valid access-token all the time; just get one when needed.
Btw, it seems there is still an issue with the mqtt refesh-token, it seems the refresh-token is only valid for 7 days and I'm not sure how to get a new one in time. It seems refresh_mqtt_token is just getting a new access-token and no new refresh-token :-(

Issue #133, opened 2025-04-08
Token expiration


andreadegiovine:
### Mark

Opel

### Model

Mokka-e

### Year

2022

### Engine

Electric

### Remote service

E-remote

### What happened?

@MoellerDi yestartday i update the integration on my hass and get some errors on reauth flow, it seem an error on expiration management (see log attached).
For the moment i install version 0.3.*

What happens?

### Log output

```shell

```

andreadegiovine:
[home-assistant.log](https://github.com/user-attachments/files/19656205/home-assistant.log)

MoellerDi:
Thanks for sharing, it seems that's the bug I'm hunting for some days now.... 
It shows an expired mqtt refresh_token in action :-) 

This debug-log also shows other areas we the integration needs some love... (verification a sensor doesn't exist we create them)

Back to the bug I'm hunting...  As of now we do not have any code in the integration taking care on refreshing the mqtt refresh_token. At least don't see it somewhere. I have a PR almost ready to mitigate that bug and to store newly provided mqtt refresh_tokens in config.

Some more detailed explanation of what IMHO happed. The mess started because the mqtt refresh_token ```'ece87848-2281-44c9-b85d-********'``` was invalid. I noticed the mqtt refresh_token seems it only last 7 days or so.

```
2025-04-07 20:25:55.005 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ---------- START connect_mqtt
2025-04-07 20:25:55.005 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ---------- START refresh_mqtt_token
2025-04-07 20:25:55.005 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ------------- access_token valid until: 2025-04-07 08:14:59.895356+02:00
2025-04-07 20:25:55.005 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ---------- START make_http_request
2025-04-07 20:25:55.513 ERROR (MainThread) [custom_components.stellantis_vehicles.stellantis] POST request error 400: https://api.groupe-psa.com/connectedcar/v4/virtualkey/remoteaccess/token?client_id=07364655-93cb-4194-8158-6b035ac2c24c&locale=it-IT
[...]

PR #134, opened 2025-04-10
Store optional refresh token if available (fix issue #133)

MoellerDi:
This fix partly issue #133 (outdated mqtt refresh_token)
As of now the code didn't store new MQTT refresh_token even if available from API.


andreadegiovine:
0.4.3-beta.2

PR #146, opened 2025-04-27
Rework close_session(), verifying an open session exist before closing

MoellerDi:
I noticed that in some cases the code threw an exception for non-existent sessions

### DIFF
--- a/custom_components/stellantis_vehicles/stellantis.py
+++ b/custom_components/stellantis_vehicles/stellantis.py
@@ -71,9 +71,7 @@ class StellantisBase:
             self._session = aiohttp.ClientSession()
 
     async def close_session(self):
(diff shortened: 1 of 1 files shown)


custom_components/stellantis_vehicles/stellantis.py
```diff
@@ -71,6 +71,10 @@ def start_session(self):
             self._session = aiohttp.ClientSession()
 
     async def close_session(self):
+        if not self._session:
```
andreadegiovine, line 74: use "or" to merge this 2 if

MoellerDi: done

andreadegiovine:
0.4.3-beta.5

Issue #114, opened 2025-03-26
Reauth again needed following invalid grant , grant expired


chmtc94:
### Mark

DS

### Model

DS4 E-Tense

### Year

2025

### Engine

Hybrid

### Remote service

E-remote

### What happened?

Same issue today as #95 already reported last week and closed.

I join the log file after improvement of my script log_anonymizer to be able to distinguish all modifications of access tokens.
You will also find in it new debug lines for easier visibility of token expiration dates 

It is to be noticed that last invalid grant was fired on 19/03/2025 09H53. That is almost exactly equal to 7 days plus one hour.

Could it be that the expiration of the authorization is the consequence of exceeding a possible quota of the number of requests per week?

Hope this help.

As I don't need to use my car for some days, I wait a while before restarting the authentication procedure.

Any suggestion appreciate...

![Image](https://github.com/user-attachments/assets/a0c42492-e149-4c85-88f6-983e3cc64073)

[2025-03-26 10-52-14. - INVALID GRANT NEW ISSUE_anonymized.log](https://github.com/user-attachments/files/19473421/2025-03-26.10-52-14.-.INVALID.GRANT.NEW.ISSUE_anonymized.log)

### Log output

```shell

```

QualQuek:
Same for me, i've already deleted and reassigned my phone into MyPeugeot App because of this just yesterday...
I'm tired to auth again and again and again!

chmtc94:
From the log file transmitted above, I made a PDF file (attached) with clarify to me the events sequence.

I had an update interval set to 300s because, not using my vehicle and not charging it, I consider not useful to ask for updates which return always the same datas. May be setting this interval to a greater value when vehicle is not used or charging, has to be considered

The color in the attached PDF correspond to routines that were called each update interval.

- yellow : update_data (update entities from vehicle datas returned by get_vehicle_status)
- green : get_vehicle_status (send request to stellantis server to get vehicle datas)
- magenta : refresh_token (refresh access_token for get_vehicle_status request if token is about to expire)
- cyan : refresh_mqtt_token (refresh mqtt access_token for commands via mqtt requests if token is about to expire)

What i observe in the logs:

- All new requested access_token has an expiry time of one hour (minus 1 sec)
- All new requested mqtt_access_token has an expiry time of 15 min (minus 1 sec)   
- access_token refresh requests are repeatedly sent some minutes before expiration
[...]

PR #147, opened 2025-04-27
Refresh_mqtt_token() revised, otp.bin file used to get new mqtt refresh_token

MoellerDi:
This is another measure to resolve the error {'error_description': 'grant is invalid', 'error': 'invalid_grant'}, which is usually caused by an expired MQTT refresh token. Unlike the token refresh for the https-based API, the API does not provide a new MQTT refresh token when refreshing an access token.

In addition (and as described at https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/issues/114#issuecomment-2823821063), the MQTT API requires the PIN code to be entered every 7 days.

To work around this, we store the OTP status in an otp.bin file during the initial authentication and reuse it later (currently every 3 days) to request a new MQTT refresh token. Similar to how [psa_car_controller](https://github.com/flobz/psa_car_controller) is doing it.

**It's expected that authentication will be required almost immediately after the update, as the required otp.bin file needs to be created, and therefore authentication is required. Hopefully, this will be the last re-authentication for a long time.**

As usual, this PR has been tested on my HA instance and works for me, but further testing may be required.
@andreadegiovine: if possible can you please put this into a dedicated beta-release?

### DIFF
--- a/custom_components/stellantis_vehicles/otp/otp.py
+++ b/custom_components/stellantis_vehicles/otp/otp.py
@@ -15,7 +15,7 @@ from . import oaep
 from .load import IWData
 
 # pylint: disable=invalid-name
-CONFIG_NAME = "otp.bin"
+OTP_FILE_NAME = "otp.bin"
 TIMEOUT_IN_S = 10
 
 logger = logging.getLogger(__name__)
@@ -126,7 +126,7 @@ class Otp:
 
         R0 = self.challenge + ";" + iw + ";" + self.get_serial()
         R1 = self.challenge + ";" + iw + ";" + self.data.iwK1
-        logger.debug("%s\n%s\n%s", R0, R1, R2)
+        #logger.debug("%s\n%s\n%s", R0, R1, R2)
         return {"R0": hashlib.sha256(R0.encode("utf-8")).hexdigest(),
                 "R1": hashlib.sha256(R1.encode("utf-8")).hexdigest(),
                 "R2": hashlib.sha256(R2.encode("utf-8")).hexdigest()}
@@ -149,7 +149,7 @@ class Otp:
             mini = x * 128
             ciphertext = cipher.decrypt(enc_b[mini:maxi])
             dec_string += ciphertext.hex()
-        logger.debug(dec_string)
+        #logger.debug(dec_string)
         return dec_string
 
     def request(self, param, setup=False):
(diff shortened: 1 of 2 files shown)


andreadegiovine:
@MoellerDi good job, please revert the edits from /otp path (leave this files as original from psa_controller) and save otp data on config (like tokens)

MoellerDi:
I still need to work on this, so I tune it to draft

andreadegiovine:
> I still need to work on this, so I make it a draft

I dont remember what data use psa_controller after saving otp object to request new token, but we can store only this and use it again (or we can store only data to load other object like original)

MoellerDi:
> > I still need to work on this, so I make it a draft
> 
> I dont remember what data use psa_controller after saving otp object to request new token, but we can store only this and use it again (or we can store only data to load other object like original)

I checked the source code of psa_controller and as far as I can tell, it does the same, [load_otp](https://github.com/flobz/psa_car_controller/blob/d8223c4bb34b24e59123c462393a4f91881ac9f3/psa_car_controller/psa/RemoteClient.py#L105), [get_otp_code](https://github.com/flobz/psa_car_controller/blob/d8223c4bb34b24e59123c462393a4f91881ac9f3/psa_car_controller/psa/RemoteClient.py#L195) and [save_otp](https://github.com/flobz/psa_car_controller/blob/d8223c4bb34b24e59123c462393a4f91881ac9f3/psa_car_controller/psa/RemoteClient.py#L200C9-L200C17).

If I understand correctly, the OTP class (otp.py) maintains a kind of OTP session with the Stellantis backend using rolling secrets/keys; in doing so, it opens multiple HTTPS requests to https://otp.mpsa.com. We would (in my opinion) have to completely parse and rewrite otp.py to extract and reload the relevant secrets/keys.....
I'm not saying it's not possible, but in my opinion there's a reason why psa_controller uses the save_otp/load_otp approach.

MoellerDi:
Okay, this PR is now stable and finally works on my HA setup. I'll run further tests to see if it maintains a valid refresh token for more than 7 days. The problem was that the first OTP authentication attempt failed for some reason, but the second one succeeded. So I added a two-attempt logic.

@andreadegiovine, can you please review this PR and give me your feedback?

andreadegiovine:
0.4.3-beta.5

andreadegiovine:
@MoellerDi please to save otp file use same path as for vehicle image

Issue #80, opened 2025-03-16
Problem with button for preconditioning


Jordan87:
### Mark

Peugeot

### Model

308 PlugInHybrid

### Year

2023

### Engine

Hybrid

### Remote service

E-remote

### What happened?

I found some issue with button for preconditionig.
 i recieve message "Failed to perform the action button/press. argument of type 'NoneType' is not iterable"

![Image](https://github.com/user-attachments/assets/ae11b551-9aff-4aa2-9bc0-fd46f36019dd)

### Log output

```shell
2025-03-16 10:49:04.719 ERROR (MainThread) [homeassistant.components.websocket_api.http.connection] [139774344380480] Unexpected exception
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/components/websocket_api/commands.py", line 245, in handle_call_service
    response = await hass.services.async_call(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<7 lines>...
    )
    ^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2794, in async_call
    response_data = await coro
                    ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2837, in _execute_service
    return await target(service_call)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 1007, in entity_service_call
    single_response = await _handle_entity_call(
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^
        hass, entity, func, data, call.context
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 1079, in _handle_entity_call
    result = await task
             ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/button/__init__.py", line 132, in _async_press_action
    await self.async_press()
  File "/config/custom_components/stellantis_vehicles/button.py", line 136, in async_press
    await self._coordinator.send_air_conditioning_command(self.name)
  File "/config/custom_components/stellantis_vehicles/base.py", line 164, in send_air_conditioning_command
    await self.send_command(button_name, "/ThermalPrecond", {"asap": new_status, "programs": self.get_programs()})
                                                                                             ~~~~~~~~~~~~~~~~~^^
[...]

PR #149, opened 2025-04-28
Handle case where `program["occurence"]` is `None`

khenderick:
I has the same error as reported via https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/issues/80#issuecomment-2745174076, which prevented me from starting/stopping the airco remotely.

This change fixes this issue by handling the case where `program["occurence"]` is `None`. I've validated the changes on my own setup/car and can now start and stop the airco.

### DIFF
--- a/custom_components/stellantis_vehicles/base.py
+++ b/custom_components/stellantis_vehicles/base.py
@@ -139,23 +139,24 @@ class StellantisVehicleCoordinator(DataUpdateCoordinator):
             if current_programs:
                 for program in current_programs:
                     if program:
-                        if program and program.get("occurence", {}).get("day", None) and program.get("start", None):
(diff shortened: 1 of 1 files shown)


khenderick:
Requested tests to execute:
* [ ] New installation
* [ ] Reconfiguration
* [ ] Sensors
  * The code is only executed on airco start/stop button press, so the above 3 seemed irrelevant.
* [x] Commands
  * I've tested the airco start/stop button for both the starting and the stopping
* [ ] 1 week without errors in logs
  * I've only done a manual validation today, but I'll keep on monitoring the logs :+1:.

andreadegiovine:
0.4.3-beta.5

Issue #130, opened 2025-04-06
Dev-branch is out-of-sync (behind master)


MoellerDi:
### Mark

Citroen

### Model

eC4

### Year

2024

### Engine

Electric

### Remote service

Connect Plus

### What happened?

@andreadegiovine the idea seems to be to actively develop in the new dev branch and occasionally merge dev into master branch. So, the dev branch should IMHO be ahead of the master branch, not behind ;-)

### Log output

```shell

```

Jordan87:
How it will work from HACS installation, will be dev-branch visible in version list or manual installation (file replacement) will be required?

MoellerDi:
Well, if I got it right one can release from dev-branch and mark them as "pre-release". That way it will show up in HACS as below. As you see if I would not have selected another version 0.3.20 (in my fork the latest release) would be downloaded. However, 0.4.0-pre as pre-release can be selected.
For better/easier PR handling the dev-branch would be the default-branch.... pre-releases pushed from dev-branch; normal-releases from main/master-branch. Should be easy; pre-releases could even be automated with github-actions.

<img width="572" alt="Image" src="https://github.com/user-attachments/assets/6ac2d8b3-f47f-4dfd-af42-1be805a2d9c5" />


IMHO key it to use proper tags and the right brach while drafting the release....

<img width="727" alt="Image" src="https://github.com/user-attachments/assets/ae3b4853-116b-4677-857f-edb7201ad4b3" />

<img width="448" alt="Image" src="https://github.com/user-attachments/assets/aab0e571-529a-4eb8-a8ee-b837a67be6d6" />

andreadegiovine:
@MoellerDi 

- develop branch updated
- develop branch is now default branch
- set workflow to pre-release on develop PR

PR #153, opened 2025-05-06
Rework refresh_token logic (refresh_token needs be called earlier)

MoellerDi:
I noticed that the 401 error might be caused by an expired access token (used as a bearer token in the authorization header). Since ```refresh_token``` in ```make_http_request``` is called before the actual HTTP request, but after we apply the query parameters and headers, the integration might run into a situation where the token is refreshed but only used in the next HTTP request and not applied in the current HTTP request. In my opinion, the best solution is to move the ```refresh_token``` calls to ```StellantisVehicleCoordinator _async_update_data```, just before ```get_vehicle_status```, as this ensures it is called at the beginning of each update interval for each vehicle.
This PR is tested on my HA system and works for me. However, might need more testing, so it would be good if you could make another pre-release.

### DIFF
--- a/custom_components/stellantis_vehicles/base.py
+++ b/custom_components/stellantis_vehicles/base.py
@@ -47,6 +47,8 @@ class StellantisVehicleCoordinator(DataUpdateCoordinator):
     async def _async_update_data(self):
         _LOGGER.debug("---------- START _async_update_data")
         try:
+            # Update token
+            await self._stellantis.refresh_token()
             # Vehicle status
             self._data = await self._stellantis.get_vehicle_status(self._vehicle)
         except ConfigEntryAuthFailed as e:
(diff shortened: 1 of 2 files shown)


andreadegiovine:
0.4.3-beta.6

spcqike:
Is there a way to test this versions? I don’t get an update notification in HA for all the beta releases. But I would love to test this as the 7 day auto logout is …. Not good (:

andreadegiovine:
@MoellerDi new release is out.

Next step:
- add permanent notification on some errors (like no vehicles found);
- delete otp file on integration uninstalling (i not remeber if vehicle images will deleted);

Unforunally i cant test nothing right now because my vehicle get errors on all command oon the last 5 months (no internet connection) and Opel support cant help me.

MoellerDi:
> Is there a way to test this versions? I don’t get an update notification in HA for all the beta releases. But I would love to test this as the 7 day auto logout is …. Not good (:

@spcqike yes, see https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/issues/130#issuecomment-2781652482 for details on how to install a pre-release. However meanwhile [0.4.3](https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/releases/tag/0.4.3) has landed,  so better install that one.

Issue #155, opened 2025-05-07
V0.4.3 has error when pressing Wake up button:  'NoneType' object has no attribute 'publish'


Mike-cpu404:
### Mark

Citroen

### Model

C5

### Year

2021

### Engine

Hybrid

### Remote service

E-remote

### What happened?

I have update to v 0.4.3 and also re-auth'd mt vehicle as it expired due to the old 7 day issue from previous versions.
However now when I press WAKE UP I am getting the error:  'NoneType' object has no attribute 'publish'

Here are part of the logs:

`2025-05-07 12:58:25.488 ERROR (MainThread) [custom_components.stellantis_vehicles.stellantis] Unexpected error during MQTT message sending: 'NoneType' object has no attribute 'publish'
2025-05-07 12:58:25.488 ERROR (MainThread) [custom_components.stellantis_vehicles.base] Failed to send command WakeUp: 'NoneType' object has no attribute 'publish'
2025-05-07 12:58:25.488 ERROR (MainThread) [homeassistant.components.websocket_api.http.connection] [281472864150304] Unexpected exception
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/components/websocket_api/commands.py", line 245, in handle_call_service
    response = await hass.services.async_call(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<7 lines>...
    )
    ^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2794, in async_call
    response_data = await coro
                    ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/core.py", line 2837, in _execute_service
    return await target(service_call)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 1007, in entity_service_call
    single_response = await _handle_entity_call(
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^
        hass, entity, func, data, call.context
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/src/homeassistant/homeassistant/helpers/service.py", line 1079, in _handle_entity_call
    result = await task
             ^^^^^^^^^^
  File "/usr/src/homeassistant/homeassistant/components/button/__init__.py", line 132, in _async_press_action
    await self.async_press()
[...]

PR #157, opened 2025-05-08
Add MQTT connection check to button state validation

MoellerDi:
This PR adds an MQTT connection check to the StellantisBaseButton class to ensure that MQTT is actually connected before exposing buttons. This is intended to prevent users from pressing a button while MQTT is not connected, since the button/MQTT command cannot be sent anyway (see #155).

As usual, this PR has been tested on my system and works for me.


khenderick (approved):

andreadegiovine:
What happens if mqtt connection is off?
The buttons remain unavailable

Have more sense to check if is off and try reconnection before send mqtt action to server? If reconnection fail we can show a persistant notify to advise user

MoellerDi:
@andreadegiovine I agree with you, so I will close this PR for now

PR #160, opened 2025-05-09
Implement async_remove_entry to clean up resources on entry removal

MoellerDi:
This PR implements ```async_remove_entry```. When the user uninstalls this integration, the OTP file, vehicle images, and image folder (if empty) are cleaned up. Obsolete repairs are also removed. Furthermore, OTP files and images are migrated to the new structure. Since the ```customer_id``` is now used as the ```unique_id```, this PR should also enable multiple accounts (not tested).
As usual, this PR has been tested on my HA system and works for me.

### DIFF
--- a/custom_components/stellantis_vehicles/__init__.py
+++ b/custom_components/stellantis_vehicles/__init__.py
@@ -13,7 +13,7 @@ from .const import (
     DOMAIN,
     PLATFORMS,
     IMAGE_PATH,
-    OTP_FILE_NAME
+    OTP_FILENAME
 )
 
 _LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


custom_components/stellantis_vehicles/__init__.py
```diff
@@ -56,4 +61,40 @@ async def async_unload_entry(hass: HomeAssistant, config: ConfigEntry) -> bool:
...
+        # Curently just a single file, but might be changed in the future
+        # to allow for multiple OTP files
+        # (e.g. for multiple Stellantis accounts)
+        otp_filename = os.path.join(hass.config.config_dir, OTP_FILE_NAME)
+        if os.path.isfile(otp_filename):
+            _LOGGER.debug(f"Deleting OTP-File: {otp_filename}")
+            os.remove(otp_filename)
+
+        # Remove related vehicle images
+        if "vehicles" in config.data:
```
andreadegiovine, line 88: We can get vehicles using same logic used on entity (https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/blob/a57a522e2bccc672db72c92d4c09e8bec09120e2/custom_components/stellantis_vehicles/button.py#L20) to load stellantis instance and the method “get_user_vehicles()”.

In this way we can reduce the config_entry file size.

Can you check and try?

Great job

andreadegiovine: Ok, if we use a directory using entry_id for images and otp file?
Like:
```
.storage/stellantis_vehicles/ENTRYID**********/otp.pickle
www/stellantis_vehicles/ENTRYID**********/VEHICLE_VIN.png
```
With this approach we can delete all file under entry_id directories and improve the multi-account OTP management

What do you think?

MoellerDi: @andreadegiovine: done, kindly review again

andreadegiovine:
https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/releases/tag/0.4.4-beta.4

custom_components/stellantis_vehicles/__init__.py
```diff
@@ -22,6 +27,23 @@ async def async_setup_entry(hass: HomeAssistant, config: ConfigEntry):
     hass.data.setdefault(DOMAIN, {})
     hass.data[DOMAIN][config.entry_id] = stellantis
 
+    # Migrate to new file structure - can be removed in future versions
```
andreadegiovine, line 30: @MoellerDi what you think about "async_migrate_entry" method? Can we use this to manage config structure (files in this case) on updates?

MoellerDi: @andreadegiovine good idea, I'll take a look

PR #210, opened 2025-06-25
Fix: make_http_request - clear result on 404 error to avoid returning error details as result.

MoellerDi:
This PR fixes a potential issue by avoiding the return of error details.
As usual, this PR was briefly tested on my HA system and works fine.


andreadegiovine:
Included on last prerelease

PR #214, opened 2025-06-27
Add grace period to refresh_token refresh logic

MoellerDi:
This PR adds a grace period to the refresh token refresh logic.

I noticed that sometimes the first and even second attempt failed, but the OTP code worked perfectly after restarting the integration (no reauthentication required). Apparently, the OTP code worked perfectly even weeks later, while HA wasn't running at all (I was on vacation for a few weeks and far away from my development system, but the OTP code worked perfectly as soon as I started it upon my return).

In short, this PR changes the refresh logic so that the refresh token continues to be refreshed with OTP code on every function call (at least every 12 hours when the "scheduled_press" function is called).

However, a fallback is now performed, where the access token is updated with a still valid refresh token, and the "ConfigEntryAuthFailed" exception is not thrown after a second failed attempt, as was previously the case.
The "ConfigEntryAuthFailed" exception is only thrown if the refresh token has expired.

Since MQTT_REFRESH_TOKEN_TTL (in const.py) is set to 3 days and the refresh token is intended to be valid for up to 7 days, there is sufficient time (up to 8 scheduled press cycles) to refresh both the refresh token and the access token.

As usual, this PR has been tested on my HA system and works for me.

### DIFF
--- a/custom_components/stellantis_vehicles/stellantis.py
+++ b/custom_components/stellantis_vehicles/stellantis.py
@@ -478,30 +478,31 @@ class StellantisVehicles(StellantisOauth):
                     await self._update_mqtt_config(mqtt_config, token_request)
             except ConfigEntryAuthFailed as e:
                 try:
-                    _LOGGER.debug("------------- 1st attempt to refresh MQTT token failed, trying again")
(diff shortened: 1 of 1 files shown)


andreadegiovine:
2025.6.7-beta.4

Issue #219, opened 2025-07-01
Battery status not updating


FrankTub:
### Mark

Citroen

### Model

E-C4

### Year

2022

### Engine

Electric

### Remote service

E-remote control

### HomeAssistant version

2025.6.1

### Integration version

2025.6.7-beta.4

### What happened?

I'm having issues on one of the latest versions where some of the sensors are no longer updated. This used to work fine.

For example my battery percentage graph looks like:

<img width="557" alt="Image" src="https://github.com/user-attachments/assets/03fabd49-ded0-41f2-a673-0de44b174ace" />

While my battery capacity looks like:

<img width="573" alt="Image" src="https://github.com/user-attachments/assets/386e0d0a-98b8-4f75-a812-9f163ca5c75e" />

There are also some errors on connectivity again, which I haven't had for quite some time now.

Some lines of the log:

```log
2025-07-01 04:37:43.399 ERROR (MainThread) [custom_components.stellantis_vehicles.stellantis] Unexpected error during HTTP request: Cannot connect to host api.groupe-psa.com:443 ssl:default [Connection reset by peer]
2025-07-01 04:37:43.400 ERROR (MainThread) [custom_components.stellantis_vehicles.base] Cannot connect to host api.groupe-psa.com:443 ssl:default [Connection reset by peer]
```

And some more:

### Log output

```shell
2025-07-01 10:09:05.006 DEBUG (paho-mqtt-client-) [custom_components.stellantis_vehicles.stellantis] Sending PINGREQ
2025-07-01 10:09:05.038 DEBUG (paho-mqtt-client-) [custom_components.stellantis_vehicles.stellantis] Received PINGRESP
2025-07-01 10:09:08.140 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ---------- START send_mqtt_message
2025-07-01 10:09:08.140 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ---------- START refresh_mqtt_token
2025-07-01 10:09:08.140 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ------------- mqtt access_token valid until: 2025-06-30 22:24:07.139297+02:00
2025-07-01 10:09:08.140 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] https://api.groupe-psa.com/connectedcar/v4/virtualkey/remoteaccess/token?client_id=<client_id>&locale=nl-NL
[...]

Issue #211, opened 2025-06-26
feature unlock/lock entity


riddik14:
### Mark

Peugeot

### Model

2008

### Year

2021

### Engine

Fuel

### Remote service

E-remote control

### HomeAssistant version

2026.6.3

### Integration version

ultima stable

### What happened?

I noticed that the integration offers the door status sensor and two open and close buttons, to simplify I created a lock template: do you think it can be integrated directly into the code? I also noticed that the intake air temperature and coolant liquid read 215°, I have not tried with the engine running but could it be an error and therefore adequate with this correction? 

 
``
- platform: template
name: "Portiere Auto"
value_template: "{{ is_state('binary_sensor.vin_portiere', 'off') }}"
lock:
service: button.press
target:
entity_id: button.vin_chiudi_portiere
unlock:
service: button.press
target:
entity_id: button.vin_apri_portiere
icon_template: >-
{% if is_state('binary_sensor.vr3ushnskmj530338_portiere', 'on') %}
mdi:car-door
{% else %}
mdi:car-door-lock
{% endif %}
``
``
"value_map": {
"path": ["engines", 0, "extension", "thermic", "coolant", "temp"],
"transform": lambda x: x / 10
}
``
2025-06-26 10:22:48.118 DEBUG (MainThread) [custom_components.stellantis_vehicles.base] ---------- START _async_update_data
2025-06-26 10:22:48.118 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ---------- START refresh_token
2025-06-26 10:22:48.118 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ------------- access_token valid until: 2025-06-26 10:51:26.578249+02:00
2025-06-26 10:22:48.118 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ---------- END refresh_token
2025-06-26 10:22:48.118 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ---------- START get_vehicle_status
2025-06-26 10:22:48.119 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ---------- START make_http_request
2025-06-26 10:22:48.253 DEBUG (paho-mqtt-client-) [custom_components.stellantis_vehicles.stellantis] Sending PINGREQ
[...]

PR #221, opened 2025-07-02
Fix: Improve value_was_updated logic to handle unchanged updated_at

adonix31:
Proposed fix to issue #219

### DIFF
--- a/custom_components/stellantis_vehicles/base.py
+++ b/custom_components/stellantis_vehicles/base.py
@@ -368,18 +368,20 @@ class StellantisBaseEntity(CoordinatorEntity):
             if self._key == "battery_soh":
                 self._value_map[6] = "capacity"
 
-    def value_was_updated(self, compare_value = False):
+    def value_was_updated(self, compare_value: bool = False):
         """ Check if value was changed. """
-        current_updated_at = None
-        if self._attr_extra_state_attributes.get("updated_at"):
-            current_updated_at = self._attr_extra_state_attributes.get("updated_at")
+        current_updated_at = self._attr_extra_state_attributes.get("updated_at")
         new_updated_at = self.get_updated_at_from_map(self._updated_at_map)
-        if compare_value:
-            current_value = self._coordinator._sensors.get(self._key)
-            new_value = self.get_value(self._value_map)
-            return current_updated_at != new_updated_at or current_value != new_value
-        else:
-            return current_updated_at != new_updated_at
+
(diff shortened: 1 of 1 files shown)


adonix31:
Now avoids unnecessary updates to battery_charging_end caused by recalculating datetime

andreadegiovine:
@adonix31 thanlks for your support, for the moment please try the last beta, the logic from this pr is similar the logic before this improvment, if last beta dont fix the issue or the charging end sensor is not stable, we can work starting from this pr to a more efficient solution

adonix31:
@andreadegiovine , thank you for your support. I undertand. Currently i am trying beta4, we'll see.

On another topic, i noticed that temperature sensors of my hybrid are being devided by 10. Values for example are now 8.9deg instead to 89deg.
This seem to be computed in get_value, but why the division? Is it needed for other car types?
```
        if key in ["coolant_temperature", "oil_temperature", "air_temperature"]:
            value = float(value) / 10
```

andreadegiovine:
> @andreadegiovine , thank you for your support. I undertand. Currently i am trying beta4, we'll see.
> 
> On another topic, i noticed that temperature sensors of my hybrid are being devided by 10. Values for example are now 8.9deg instead to 89deg. This seem to be computed in get_value, but why the division? Is it needed for other car types?
> 
> ```
>         if key in ["coolant_temperature", "oil_temperature", "air_temperature"]:
>             value = float(value) / 10
> ```

My vehicle has not this data, this is a request from other user issue

adonix31:
What vehicle type he has, because clearly, for my hybrid, this division is incorrect.

andreadegiovine:
https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/issues/211

chmtc94:
see #225

riddik14:
forse cosi è meglio. il valore oil temp che esiste veniva fuori 8,3°C ed è essato. di ho aperto una pull per questo con questa modifica      


  if key in ["battery_capacity", "battery_residual"]:
            if int(value) < 1:
                value = None
            else:
                value = (float(value) / 1000) + 10

        if key in ["coolant_temperature", "air_temperature"]:
            value = float(value) / 10
        
        if key in ["oil_temperature"]:
            value = float(value) 

        if isinstance(value, str):
            value = value.lower()

        return value

andreadegiovine:
@riddik14 https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/pull/224

PR #259, opened 2025-08-12
Added daylight sensor entity

vidarino:
Added parsing of daylight sensor present in e.g. Opel Mokka state updates:

```
{
  "environment": {
    "luminosity": {
      "createdAt": "2025-08-12T07:06:34Z",
      "day": true
    },
    "air": {
      "createdAt": "2025-08-12T07:06:34Z",
      "temp": 13.5
    }
  },
  ...
}
```


andreadegiovine:
Hi @vidarino can you share a use case for this information on hass?

vidarino:
Hi. It's a bit of a fringe case, perhaps, but it's one more light sensor. :)

I've used it to see if the car is in the garage, and as a trigger when I pass through a tunnel near my home (later replaced with zone info, but it's possible, at least).

Issue #260, opened 2025-08-13
Errors in TOKEN renewal during automatic wake-up (every 12 hours after the last HA restart)


chmtc94:
### Mark

DS

### Model

DS4 E-Tense

### Year

2022

### Engine

Hybrid

### Remote service

E-remote control

### HomeAssistant version

2025.8.1

### Integration version

2025.8.2-beta.2

### What happened?

@MoellerDi  @andreadegiovine 

2025-08-13 15H49 : Errors in TOKEN renewal during automatic wake-up (every 12 hours after the last HA restart).

I'm not confident enough with the procedure to properly investigate. I wonder on the timing of all events and actions... Perhaps some actions/commands with new token execute too soon after token renewal ?

I provide the corresponding logs to help to confirm or improve error management...
In the log file all tokens are numbered with a different number for each different token (a same number is the same token everywhere it appears).

NB : Despite the errors, everythings seems OK again at the end of the sequence. I executed many trips after the sequence with no apparent issue and a charging sequence is programmed that night. Will see tomorrow.

Hope this could help

[2025-08-13 - 15H49 - Error in TOKEN renewal during automatic wake-up.log](https://github.com/user-attachments/files/21762559/2025-08-13.-.15H49.-.Error.in.TOKEN.renewal.during.automatic.wake-up.log)

### Log output

```shell

```

chmtc94:
@MoellerDi 

It seems to me that, since we're careful to request renewal of access tokens and refresh tokens before their expected expiration dates, we shouldn't encounter these errors.

My feeling is that (not at all sure I'm right) we're facing an interference conflict between the two sequences:
- the periodic refresh of vehicle data every TIME_INTERVAL period
- the periodic vehicle wake-up command (every 12 hours after the last HA restart), but which could also be any other manual command sent by clicking a command button during a periodic vehicle data update.

Any token refreshes in one sequence seem to disrupt the other sequence when the two sequences interfere. Perhaps it would be necessary to prevent this interference by not starting one of them until the other is completely finished (token refresh(es) - command - response - update )? 

NB: This is not my case, but the problem could perhaps be even more critical if managing several vehicles.

Hope this will help...

UPDATE : 
Everything however still OK until now. Charging still in progress at the moment. 
As no automatic data are sent by my vehicle during charging (as usual with my vehicle) I tested a manual wake-up via wake-up button to force update. That was fully successful and vehicle data were well updated a while after the command. 
[...]

Issue #251, opened 2025-08-05
MQTT connection fails (reset by peer), while other API queries work


dbeinder:
### Mark

Peugeot

### Model

e-208

### Year

2021

### Engine

Electric

### Remote service

E-remote control

### HomeAssistant version

2025.7.4

### Integration version

2025.7.5-beta.4

### What happened?

The MQTT connection fails to connect with `ConnectionResetError: [Errno 104] Connection reset by peer`, however all other queries seem to succeed without errors. The integration worked fully previously - then there were some timeout errors tonight. I restarted HA with debug logging. Unfortunately I have lost the earlier logs. The log below shows the behavior after HA restart.

### Log output

```shell
025-08-05 08:48:20.381 WARNING (SyncWorker_0) [homeassistant.loader] We found a custom integration stellantis_vehicles which has not been tested by Home Assistant. This component might cause stability problems, be sure to disable it if you experience issues with Home Assistant
2025-08-05 08:48:20.385 WARNING (SyncWorker_0) [homeassistant.loader] We found a custom integration hacs which has not been tested by Home Assistant. This component might cause stability problems, be sure to disable it if you experience issues with Home Assistant
2025-08-05 08:48:20.408 WARNING (SyncWorker_0) [homeassistant.loader] We found a custom integration tuya_local which has not been tested by Home Assistant. This component might cause stability problems, be sure to disable it if you experience issues with Home Assistant
2025-08-05 08:48:44.791 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] {'access_token': '1df3bb51-5ef4-4ade-a262-QQQQQQQQ', 'country_code': 'AT', 'customer_id': 'AP-ACNT200013QQQQQ', 'expires_in': '2025-08-05T08:49:25.041095+02:00', 'mobile_app': 'MyPeugeot', 'oauth_url': 'https://idpcvs.peugeot.com', 'realm': 'clientsB2CPeugeot', 'scheme': 'mymap', 'locale': 'de-AT', 'client_id': '1eebc2d5-5df3-459b-a624-QQQQQQQQ', 'client_secret': 'QQQQQQQQQQQQQQQQQQQ', 'basic_token': 'QQQQQQQQQQQ0MwbEEyaUUyYVI3Z0s2dUU1ckYzbEo4cEMzbk8xcFI3dEw4dlUx', 'culture': 'at'}
[...]

PR #261, opened 2025-08-14
Revise token refresh logging to make it clearer and avoid confusion - No functional changes

MoellerDi:
As described in #260, the current logging can be misleading and confusing because it logs things as "errors" even if the "errors" are handled later in the calling function.

### DIFF
--- a/custom_components/stellantis_vehicles/stellantis.py
+++ b/custom_components/stellantis_vehicles/stellantis.py
@@ -139,7 +139,7 @@ class StellantisBase:
                 if method != "DELETE" and (await resp.text()):
                     result = await resp.json()
                 if not str(resp.status).startswith("20"):
-                    _LOGGER.error(f"{method} request error {str(resp.status)}: {resp.url}")
+                    _LOGGER.debug(f"{method} request error {str(resp.status)}: {resp.url}")
                     _LOGGER.debug(headers)
                     _LOGGER.debug(params)
                     _LOGGER.debug(json)
@@ -167,7 +167,8 @@ class StellantisBase:
                 _LOGGER.debug("---------- END make_http_request")
                 return result
         except ConfigEntryAuthFailed as e:
-            _LOGGER.error("Authentication failed during HTTP request: %s", e)
+            # Logged only as a warning, the exception could be handled by the calling function
+            _LOGGER.warning("Authentication failed during HTTP request: %s", e)
             raise
         except Exception as e:
             _LOGGER.error("Unexpected error during HTTP request: %s", e)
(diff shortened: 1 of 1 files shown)


chmtc94:
> As described in #260, the current logging can be misleading and confusing because it logs things as "errors" even if the "errors" are handled later in the calling function.
> 
> 

I tend to agree:
The error encountered should be considered as:
- a warning when the integration's processing restores proper operation, because the error is real and abnormal given that the integration ensures that all tokens are renewed before their scheduled expiration date. It therefore makes sense to continue reporting its occurrence because its origin remains unexplained...
- a (permanent) error if the integration's processing fails to restore normal operation, which is not the case in report #260

MoellerDi:
Since this PR was not considered for either 2025.8.2 or 2025.8.3, there is clearly no need for it, therefore closed

chmtc94:
@MoellerDi  

> Since this PR was not considered for either 2025.8.2 or 2025.8.3, there is clearly no need for it, therefore closed

I don't understand why this PR wasn't addressed ☹️😢 to fix #260

It would have been better to have only : 
- a warning when the error occurs and is handled properly by the integration, so you know everything continues to work correctly,
- and an error if the handling fails and the integration's processing is no longer guaranteed...

The possible alternative, which would consist of replacing the warning with a simple debug message (only visible in log file and debug mode), would have the disadvantage of not being informed of the existence and frequency of these errors which are nevertheless abnormal even if they remain without impact on the proper functioning of the integration when they are correctly handled...

In any case, leaving the code as is does not seem acceptable to me because the error appearing in the HA log casts doubt on the proper functioning of the integration, which is not the case when the processing resolves the problem (e.g., in issue #260).

MoellerDi:
> In any case, leaving the code as is does not seem acceptable to me because the error appearing in the HA log casts doubt on the proper functioning of the integration, which is not the case when the processing resolves the problem (e.g., in issue #260).

I completely agree with you, but I haven't received any feedback or comments from @andreadegiovine on this PR, so I'm not sure what to do next.

andreadegiovine:
Reopen this PR and i can merge it.

OT
After this error on my hass:
IMG_2615
all action buttons was disabled until reload integration, any idea?

MoellerDi:
@andreadegiovine, thanks for your feedback. PR is open again.

Regarding your error message, the following code seems to be in play:
https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/blob/0531c9f4534b1f2207d8ab14e6897f3e77c7fe34/custom_components/stellantis_vehicles/stellantis.py#L611-L621
Do you have the corresponding debug log handy?

Are you aware of https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/issues/251#issuecomment-3198524219 ? Perhaps this is also related to your error.

andreadegiovine:
No unfortunally i have disabled the debug, i think we need a reconnection logic in this case

Issue #246, opened 2025-07-30
Most Sensors unkown


starhopper1:
### Mark

Citroen

### Model

E-Berlingo

### Year

2025

### Engine

Electric

### Remote service

E-remote control

### HomeAssistant version

2025.7.4

### Integration version

Version 2025.7.4

### What happened?

Seems to be same as this issue
https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/issues/78

Some things got picked up but most don´t, in looks i can see all the stuff so i guess it can´t point the info to the sensors?

<img width="299" height="1198" alt="Image" src="https://github.com/user-attachments/assets/e8850ed0-45cf-4258-85df-e43d8b18203d" />

### Log output

```shell
2025-07-30 20:32:48.694 DEBUG (paho-mqtt-client-) [custom_components.stellantis_vehicles.stellantis] Sending PINGREQ
2025-07-30 20:32:48.738 DEBUG (paho-mqtt-client-) [custom_components.stellantis_vehicles.stellantis] Received PINGRESP
2025-07-30 20:32:49.361 DEBUG (MainThread) [custom_components.stellantis_vehicles.base] ---------- START _async_update_data
2025-07-30 20:32:49.361 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ---------- START refresh_token
2025-07-30 20:32:49.362 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ------------- access_token valid until: 2025-07-30 20:52:06.605750+02:00
2025-07-30 20:32:49.362 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ---------- END refresh_token
2025-07-30 20:32:49.362 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ---------- START get_vehicle_status
2025-07-30 20:32:49.363 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ---------- START make_http_request
2025-07-30 20:32:49.920 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ---------- END make_http_request
2025-07-30 20:32:49.920 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] https://api.groupe-psa.com/connectedcar/v4/user/vehicles/passwordpassword/status?client_id=passwordcae&locale=de-DE
[...]

Issue #137, opened 2025-04-15
Home Assistent crashes after some time when this intergration is on


handswork:
### Mark

Peugeot

### Model

E3008

### Year

2024

### Engine

Electric

### Remote service

E-remote

### What happened?

Thanks for all the work you are doing with this.

Have two cars. E3008 and 508

Since the latest upgrade (0.4.2), Home Assistent crashes after sending 2 commands to the vehicles. first goes fine, but second, crashes
Atleast the GUI crashes. and I can console to VM and type "host reboot" but no output, I need to reset the VM before it gets online again.
When I deactived the integration Home Assistent runs fine, atleast it does not crash.

And I don't have all the sensors for the full eletric car. mainly the battery sensors.

### Log output

```shell

```

andreadegiovine:
Any news on this issue?

handswork:
Still crashes when I send commands to the cars.
And still missing battery information and other different things on one of the cars.

andreadegiovine:
Please share details and log (we need cars status response from stellantis)

handswork:
It do looks like I can't replicate the crash anymore. but I can't send commands to the car either, and that was when it crashed before..
I have tried to reconfigure it several times, nothing chance anything.
But it looks like my hybrid car (Peugeot 508 sw) is working. it is the full electric (Peugeot E3008) that don't work.
See attachments for details. - unless you need other details?


![Image](https://github.com/user-attachments/assets/e230bf54-70e4-45f3-99d8-a450297110ec)

[home-assistant_stellantis_vehicles_2025-05-24T15-43-25.126Z.log](https://github.com/user-attachments/files/20427088/home-assistant_stellantis_vehicles_2025-05-24T15-43-25.126Z.log)

Issue #17, opened 2025-02-10
Sensors not updating


Mike-cpu404:
### Mark

Citroen

### Model

C5 Aircross

### Year

2021

### Engine

Hybrid

### Remote service

E-remote

### What happened?

v0.2.21 

The sensors do not appear to be updating, last data is 6 hours old.

There is an error showing in the logs:

```
Logger: custom_components.stellantis_vehicles.stellantis
Source: custom_components/stellantis_vehicles/stellantis.py:110
integration: Stellantis Vehicles (documentation, issues)
First occurred: 05:20:22 (2 occurrences)
Last logged: 10:22:18

GET request error 401: https://api.groupe-psa.com/connectedcar/v4/user/vehicles/<#####>/status?client_id=<######>&locale=en-GB
```

and

```
Logger: custom_components.stellantis_vehicles.base
Source: custom_components/stellantis_vehicles/base.py:46
integration: Stellantis Vehicles (documentation, issues)
First occurred: 05:20:22 (2 occurrences)
Last logged: 10:22:18

Unauthorized - Cannot pass the security checks that are required by the target API or operation, Enable debug headers for more details.
```

I have checked the debut logs but i'm unsure why they are not updating, maybe a token validation error?

Thanks,
Mike

### Log output

```shell
2025-02-10 11:04:24.082 DEBUG (MainThread) [custom_components.stellantis_vehicles.base] ---------- START _async_update_data
2025-02-10 11:04:24.082 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ---------- START get_vehicle_status
2025-02-10 11:04:24.083 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ---------- START refresh_token
2025-02-10 11:04:24.083 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ---------- END refresh_token
2025-02-10 11:04:24.083 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ---------- START make_http_request
2025-02-10 11:04:24.437 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] ---------- END make_http_request
2025-02-10 11:04:24.437 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] https://api.groupe-psa.com/connectedcar/v4/user/vehicles/<##########>&locale=en-GB
[...]

PR #263, opened 2025-08-17
Chmtc94 patch 246

chmtc94:
Try to fix issue #246 by replacing index in energies nodes by a one element dictionary containing a key value pair to find in the energies list elements instead of an indice.

### DIFF
--- a/custom_components/stellantis_vehicles/base.py
+++ b/custom_components/stellantis_vehicles/base.py
@@ -360,9 +360,9 @@ class StellantisBaseEntity(CoordinatorEntity):
     def update_maps_for_hybrid(self):
         """ Update value/updated_at map for hybrid vehicles. """
         if self._coordinator.vehicle_type == VEHICLE_TYPE_HYBRID:
-            if self._value_map[0] == "energies" and self._value_map[1] == 0 and not self._key.startswith("fuel"):
-                self._value_map[1] = 1
-                self._updated_at_map[1] = 1
+#            if self._value_map[0] == "energies" and self._value_map[1] == 0 and not self._key.startswith("fuel"):
+#                self._value_map[1] = 1
+#                self._updated_at_map[1] = 1
 
             if self._key == "battery_soh":
                 self._value_map[6] = "capacity"
@@ -376,25 +376,7 @@ class StellantisBaseEntity(CoordinatorEntity):
 
     def get_updated_at_from_map(self, updated_at_map):
         """ Get data updated_at from map. """
-        vehicle_data = self._coordinator._data
-        value = None
-        for key in updated_at_map:
-            if value is None: # first key in the map
-                if key in vehicle_data:
-                    value = vehicle_data[key]
-            else: # following keys in the map (value has been set with result of previous key)
-                if isinstance(key, int) or key in value: 
-                    value = value[key]
-                else: # value not an array and key not found in value
-                    value = None
-            if value is None: # Stop iteration immediately if None value encountered at this stage
-                break
-
-        # Two following lines may be useless...
-        # if value and not isinstance(value, str):
-        #     value = None
-
-        return value
+        return self.get_value_from_map(updated_at_map)
 
     def get_value_from_map(self, value_map):
         """ Get data value from map. """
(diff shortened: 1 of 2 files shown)


andreadegiovine:
@chmtc94 
Can this logic applied on 'engines' maps?

custom_components/stellantis_vehicles/base.py
```diff
@@ -405,17 +387,18 @@ def get_value_from_map(self, value_map):
                 if key in vehicle_data:
                     value = vehicle_data[key]
             else: # following keys in the map (value has been set with result of previous key)
-                if isinstance(key, int) or key in value: 
+                if isinstance(key, dict) and isinstance(value, list): # key is a dict and value a list
+                    # Use dictionnary in map as key_field, ke_value to look for in value list
```
MoellerDi, line 391: Typo: “ke_value” should be read as “key_value”?

chmtc94: yes 😉

MoellerDi (approved):

chmtc94:
@andreadegiovine 

For engines sensors, the problem is a little bit different because we have only one sensor which needed to be updated either by index 0 or by index 1 depending on the engine type.

In issue #246 we had two sensors (one for fuel and one for battery) that needed to be updated by index 0 or index 1, depending on the vehicle type, that may be not present or inverted in the vehicle_data buffer. This PR fix that.

However, we should be able to solve your problem using the same way 😉.
We only need a procedure update_engines_maps() similar to update_maps_for_hybrids() that will override the initial 0 indexes in the engines maps (according to SENSORS_DEFAULT in const.py) by a one element dictionary : {"type": "Thermic"} for thermic vehicles and {"type": "Electric"} for others. This procedure should be called the same way as update_maps_for_hybrids() using the supposed known vehicle type at this initialisation stage...
The already updated get_value_for_map(value_map) by this PR will then handle properly the good correspondance between vehicle type and corresponding "engines" node.

To my opinion, it should be better to make another issue + PR for this specific need...

MoellerDi:
It seems we currently have 3 'engines' sensors, and all of them are `{"type": "Thermic"}`. No?

```
    "coolant_temperature" : {
        "icon": "mdi:coolant-temperature",
        "unit_of_measurement" : UnitOfTemperature.CELSIUS,
        "device_class": SensorDeviceClass.TEMPERATURE,
        "state_class": SensorStateClass.MEASUREMENT,
        "value_map" : ["engines", 0, "extension", "thermic", "coolant", "temp"],
        "updated_at_map" : ["engines", 0, "createdAt"],
        "engine": [VEHICLE_TYPE_THERMIC, VEHICLE_TYPE_HYBRID]
    },
    "oil_temperature" : {
        "icon": "mdi:oil-temperature",
        "unit_of_measurement" : UnitOfTemperature.CELSIUS,
        "device_class": SensorDeviceClass.TEMPERATURE,
        "state_class": SensorStateClass.MEASUREMENT,
        "value_map" : ["engines", 0, "extension", "thermic", "oil", "temp"],
        "updated_at_map" : ["engines", 0, "createdAt"],
        "engine": [VEHICLE_TYPE_THERMIC, VEHICLE_TYPE_HYBRID]
    },
    "air_temperature" : {
        "icon": "mdi:thermometer-lines",
        "unit_of_measurement" : UnitOfTemperature.CELSIUS,
        "device_class": SensorDeviceClass.TEMPERATURE,
        "state_class": SensorStateClass.MEASUREMENT,
        "value_map" : ["engines", 0, "extension", "thermic", "air", "temp"],
        "updated_at_map" : ["engines", 0, "createdAt"],
        "engine": [VEHICLE_TYPE_THERMIC, VEHICLE_TYPE_HYBRID]
    }
```

Data from #246 
```
    "engines": [
        {
            "type": "Thermic",
            "extension": {
                "thermic": {
                    "coolant": {
                        "temp": 215.0
                    },
                    "oil": {
                        "temp": 215.0
                    },
                    "air": {
                        "temp": 215.0
                    }
                }
            },
            "createdAt": "2023-12-28T17:21:11Z"
        }
    ],
```
(similar data found in: #212 #211 #137 #78 #17 #8)

chmtc94:
> It seems we currently have 3 'engines' sensors, and all of them are `{"type": "Thermic"}`. No?
> 

Yes, but it depends on vehicle types and on vehicle_data record.

For my vehicle (hybrid) :

the temperature are sometimes present with a map similar to #246 

[2025-08-12 12h03.json](https://github.com/user-attachments/files/21824224/2025-08-12.12h03.json)

and sometimes not, with a value as a list of two elements (almost empty) with the "type" field with values "Thermic" and "Electric"

[2025-08-12.10H49.json](https://github.com/user-attachments/files/21824225/2025-08-12.10H49.json)

The difficulty is to have enough examples with different vehicle types and different Stellantis vehicle_data records for each to be able to more investigate...

@andreadegiovine 

Definitively, we need to detached from this PR these necessary further investigation on the engine temperature sensors. Moreover many of these temperature sensors are well updated with their 0 indexes. It might also be a Stellantis issue...

MoellerDi:
@chmtc94, thanks for your explanation. I was just wondering because the "engines" sensors (coolant temperature, oil temperature, and air temperature) only sound useful for thermic engines. I wouldn't expect an oil temperature sensor in an electric vehicle, for example. Also, the "value_map" currently points to the "thermic" "extension". So I assumed it was safe to use "{"type": "Thermic"}" as a selector for it.

But you might be right; that's better addressed in a separate PR. So let's get this one merged :-)

chmtc94:
@MoellerDi 

It also seems that temperatures appear in the "Engines" node of the Stellantis vehicle_data buffer, with the "type":"thermic" attribute, for fully electric vehicles (e.g. #246). Using filtering could thus allow ignoring these values, which are apparently completely erroneous, by indicating {"type":"Electric"} in the map of the corresponding sensors 😉

EDIT : As above suggested, the map should only be updated by the integration (replacement of 0 index by the dictionary filter) only once the vehicle type is known by the integration.

Issue #272, opened 2025-08-26
Battery Capacity Correction


dbeinder:
### Mark

Peugeot

### Model

e-208

### Year

2021

### Engine

Electric

### Remote service

E-remote control

### HomeAssistant version

2025.7.4

### Integration version

2025.8.3

### What happened?

Picking up on https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/issues/264

My car (https://ev-database.org/car/1168/Peugeot-e-208) also reports incorrect values over the API. Over the last few days, I have tracked the API value vs the value reported using Carscanner (OBD Dongle):
<img width="565" height="338" alt="Image" src="https://github.com/user-attachments/assets/d44dda2a-30c5-4aca-8585-8d0bcd8d474f" />

I can't see any offset / reserve, but the OBD kWh is consistently 34.4% larger than the API value.
I'm also quite confident that the OBD value is fairly accurate, as it matches well with the meter in my wallbox (minus charging inefficiency). How exactly Stellantis calculates 94.4% SOH on 40.8kWh usable capacity of an 50kWh battery is beyond me though. The integration also shows 94% SOH.

Would be interesting to see if others with incorrect API values have the same +34.4% linear relationship between API and real values. @andreadegiovine @kaysiel

### Log output

```shell

```

andreadegiovine:
@dbeinder this is the values from API of my Mokka-e 2022:

SOH 92%
SOC 68%
Capacity 32,64kWh
Residual 19,78kWh

dbeinder:
Seems pretty similar to mine. When you charge to 100%, how high does the residual go?
Do you have an OBD dongle to read the residual kWh from the BMS directly?
If it also 34.4% higher, maybe we can just add a config setting to the integration to apply the factor `1.344` to incoming values.

5kat3R:
Citroën ë-C4 2022
SOH 90%
SOC 77%
Capacity 32,06kWh
Residual 23,58kWh

andreadegiovine:
> Seems pretty similar to mine. When you charge to 100%, how high does the residual go? Do you have an OBD dongle to read the residual kWh from the BMS directly? If it also 34.4% higher, maybe we can just add a config setting to the integration to apply the factor `1.344` to incoming values.

Can you make here the calculations using my data and data from @5kat3R

dbeinder:
[...]

PR #278, opened 2025-09-04
battery_capacity correction to match battery residual

thewhale21:
apply correction to battery_capacity like battery_residual in latest update.


thewhale21:
image
In #272 the residual was updated but not capacity. this should apply to both

PR #279, opened 2025-09-04
Match Residual and Capacity value alterations

thewhale21:
Match Residual and Capacity value alterations from #272 
image
Currently residual can exceed capacity

### DIFF
--- a/custom_components/stellantis_vehicles/base.py
+++ b/custom_components/stellantis_vehicles/base.py
@@ -453,8 +453,7 @@ class StellantisBaseEntity(CoordinatorEntity):
             # https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/issues/272
             correction_on = self._coordinator._sensors.get("switch_battery_values_correction", False)
             if value and correction_on:
-                if key == "battery_residual":
-                    value = value * 1.343
+                value = value * 1.343
 
         if key in ["coolant_temperature", "oil_temperature", "air_temperature"]:
             value = float(value)


khenderick:
I'm not completely sure that this uniform applying of the correction factor is correct.

I have a 50kWh Opel Corsa from 2020, which has a real capacity of 46.3kWh. The integration reports a SoH of 87% which means 40.28kWh available capacity. For many months, the reported capacity was 40.75kWh, which is "close enough". About a month ago the reported available capacity suddenly dropped to 30.75kWh while the SoH is still reported to be 87%. 

30.75 (the current reported capacity) * 1.344 = 41.33, which again is more or less the "real" available capacity given a SoH of 87%. However, 40.75 (the previous reported capacity) * 1.344 = 54.77, which is way too high.

So I'm wondering whether separate switches for applying the corrections would make more sense.

thewhale21:
> I'm not completely sure that this uniform applying of the correction factor is correct.
> 
> I have a 50kWh Opel Corsa from 2020, which has a real capacity of 46.3kWh. The integration reports a SoH of 87% which means 40.28kWh available capacity. For many months, the reported capacity was 40.75kWh, which is "close enough". About a month ago the reported available capacity suddenly dropped to 30.75kWh while the SoH is still reported to be 87%.
> 
> 30.75 (the current reported capacity) * 1.344 = 41.33, which again is more or less the "real" available capacity given a SoH of 87%. However, 40.75 (the previous reported capacity) * 1.344 = 54.77, which is way too high.
> 
> So I'm wondering whether separate switches for applying the corrections would make more sense.

Your previous reported capacity (40.75) was with the + 10kwh adjustment that was removed a couple of patches ago. So with this gone the *1.344 would be a good addition. 

Before the 2025.9.1 both capacity and residual were 1.344 off. Now only my capacity is 1.344 off. 

Are you saying your residual was off and your capacity was fine?

khenderick:
Indeed, residual was off and capacity was fine for me previously. Now since a month they are both off, so now I would apply the correction to both, but I can imagine that some user might still have a similar setup as I had before, were only residual is off.

I have no clue on what Stellantis does with those numbers, or why.

thewhale21:
> Indeed, residual was off and capacity was fine for me previously. Now since a month they are both off, so now I would apply the correction to both, but I can imagine that some user might still have a similar setup as I had before, were only residual is off.
> 
> I have no clue on what Stellantis does with those numbers, or why.

Out of curiosity what were your previous values ?
With the +10kw I was seeing 42kwh as my capacity and my residual was getting up to 40kwh.

khenderick:
Oh, I think I see what was going on. My previous capacity was 40.75kWh like I stated, but I wasn't aware of this +10kWh and I didn't make the link that this 40.75kWh was including this offset. It indeed "suddenly" dropped to 30.75kWh when I updated the integration.

So nevermind, applying this 1.344 correction factor to both residual _and_ capacity totally makes sense. I'm sorry for all the confusion.

PR #284, opened 2025-09-05
Update Last Trip - Average speed to correct units

thewhale21:
image

0.31 miles taking 2mins 18 seconds works out to be 8.088 miles per hour or 3.62m/s

image

0.25 miles taking 1min 28 seconds works out at 10.23 miles per hour or 4.572m/s 

m/s Seems much closer to the real value than Km/h


andreadegiovine:
IMOH is better to do the conversion of the value to km/h instead change the unit of mesaurment because is an attribute not a sensor (where any user can change from UI the unit of mesaurment and hass make the conversion).
What you think?

thewhale21:
> IMOH is better to do the conversion of the value to km/h instead change the unit of mesaurment because is an attribute not a sensor (where any user can change from UI the unit of mesaurment and hass make the conversion). What you think?

Totally makes sense :) 
Are you okay to implement that? (I'm by no means a python expert)

PR #285, opened 2025-09-05
Update Last Trip - Attribute - Consumption / Consumption per 100km

thewhale21:
Update Last Trip Attributes to match the 1.343 from battery size. from #272

1000/1.343 = 746.26 

Have tested on a fork and this apears to match the kw/100km reported by the car.

### DIFF
--- a/custom_components/stellantis_vehicles/sensor.py
+++ b/custom_components/stellantis_vehicles/sensor.py
@@ -132,7 +132,11 @@ class StellantisLastTripSensor(StellantisRestoreSensor):
                 if consuption["type"] == VEHICLE_TYPE_ELECTRIC:
                     consumption_unit_of_measurement = UnitOfEnergy.KILO_WATT_HOUR
                     avg_consumption_unit_of_measurement = UnitOfEnergy.KILO_WATT_HOUR+"/100"+UnitOfLength.KILOMETERS
-                    divide = 1000
+                    correction_on = self._coordinator._sensors.get("switch_battery_values_correction", False)
+                    if correction_on:
+                        divide = 746.26
+                    else: 
+                        divide = 1000
                 else:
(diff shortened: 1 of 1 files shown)


custom_components/stellantis_vehicles/sensor.py
```diff
@@ -132,7 +132,11 @@ def coordinator_update(self):
                 if consuption["type"] == VEHICLE_TYPE_ELECTRIC:
                     consumption_unit_of_measurement = UnitOfEnergy.KILO_WATT_HOUR
                     avg_consumption_unit_of_measurement = UnitOfEnergy.KILO_WATT_HOUR+"/100"+UnitOfLength.KILOMETERS
-                    divide = 1000
+                    correction_on = self._coordinator._sensors.get("switch_battery_values_correction", False)
+                    if correction_on:
+                        divide = 746.26
+                    else:
```
andreadegiovine, line 138: Remove this and next row

thewhale21:
Sorry i do not understant the comments. Do you not like the proposed suggestion?

andreadegiovine:
> Sorry i do not understant the comments. Do you not like the proposed suggestion?

Its ok 😁
Restore the removed row 135
Remove the rows 138-139

and commit here

The else condition cost performance of the system, in this case we can valorize the value and only use the “if” condition override it

thewhale21:
> > Sorry i do not understant the comments. Do you not like the proposed suggestion?
> 
> Its ok 😁 Restore the removed row 135 Remove the rows 138-139
> 
> and commit here
> 
> The else condition cost performance of the system, in this case we can valorize the value and only use the “if” condition override it

image

I think we would need to multiply in that case. Otherwise the divide would apply twice. 
Though i think i maybe using the wrong command to multiply (Sorry lack of python knowledge again!)

PR #290, opened 2025-09-05
Update Last trip - attributes consumption

thewhale21:
Had previously assumed the divide by 1000 would also apply. However this was not the case so incrementing 0.76 by 1000 times to fix
Screenshot_20250905_213441_Home Assistant


andreadegiovine:
The correction factor of #272 is 1.343 then the value of this correction must be 1000/1.343=744.60
Why you use 746.26?

thewhale21:
> The correction factor of #272 is 1.343 then the value of this correction must be 1000/1.343=744.60 Why you use 746.26?

Ah I was rounding to 1.34 rather than 1.343 will change

Issue #296, opened 2025-09-09
Battery Values Correction does not immediately apply after integration restart


jostsalathe:
### Mark

Opel

### Model

Corsa-e

### Year

2019

### Engine

Electric

### Remote service

E-remote control

### HomeAssistant version

2025.9.1

### Integration version

2025.9.4

### What happened?

The recently added Battery Values Correction (#272) is not applied at startup of the integration but (for each affected sensor - residual and capacity as I am aware) only from when the first new value is received.

This produces glitches where, between Integration restart and the first new value received, the uncorrected values are presented in the graphs.

I guess this might have something to do with value initialization at integration startup.
But I had no time, yet, to further investigate...

Having the correction factor at all is really cool, though! <3

### Log output

```shell

```

thewhale21:
<img width="251" height="170" alt="Image" src="https://github.com/user-attachments/assets/90e63854-4610-4a7c-8a68-e72dd70e5cbc" />

I think we just need to have the button load before the sensor. 
Am just testing at the moment, but seems to load correctly.

thewhale21:
#289 @khenderick Not sure if you have a build you can test on aswell, I updated the above in cont.py and it seems to work.

jostsalathe:
Okay, so after installing integration version 2025.9.5, ther were none of the previouslly mentioned glitches. So I guess, the specific issue I reported here is now fixed =]

Thanks a lot, you people rock! <3

Issue #289, opened 2025-09-05
Last trip - Value decreasing on reload


thewhale21:
### Mark

Vauxhall

### Model

Corsa-e

### Year

2020

### Engine

Electric

### Remote service

E-remote control

### HomeAssistant version

2025.9.0

### Integration version

2025.9.2-beta1

### What happened?

When intergration is reloaded the value of last trip resets to a lower value

### Log output

```shell

```

khenderick:
I might be a bit more wide spread; I noticed that the residual capacity is not corrected the first time after reload. On subsequent updates it is corrected as it should be.

<img width="336" height="378" alt="Image" src="https://github.com/user-attachments/assets/842357be-d19a-407b-aec4-129082f3b81b" />

Maybe related to this, it looks like the total capacity sensor isn't corrected at all. Most likely because it's also only calculated at integration load (where there seems to be an issue), and on subsequent updates. Yet the total capacity almost never changes, so it won't get the corrected value since the "is value changed" happens on the raw data.

thewhale21:
This is down to the intergration loading the switch after the calculation is done, causing the check in base.py "correction_on = self._coordinator._sensors.get("switch_battery_values_correction", False)"

I removed the condition in a fork and it works as expected. We just need to figure out how to delay the calculation until the switch is created and the state restored.

thewhale21:
> ### Mark
> Vauxhall
> 
> ### Model
> Corsa-e
> 
> ### Year
> 2020
> 
> ### Engine
> Electric
> 
> ### Remote service
> E-remote control
> 
> ### HomeAssistant version
> 2025.9.0
> 
> ### Integration version
> 2025.9.2-beta1
> 
> ### What happened?
> When intergration is reloaded the value of last trip resets to a lower value
> 
> ### Log output

Found the issue!

If the sensor is set to another unit, other than KM it is being recalculated and converted again as if it were km.

khenderick:
> 
> Found the issue!
> 
> If the sensor is set to another unit, other than KM it is being recalculated and converted again as if it were km.

Hmm, I recently noticed `StellantisRestoreSensor.async_added_to_hass` where `async_get_last_state` is used instead of `async_get_last_sensor_data` to get the native value. See [HA developer docs](https://developers.home-assistant.io/docs/core/entity/sensor/#restoring-sensor-states). I'm not completely sure though and it probably needs some more thought, but it might be in the right direction.

PR #299, opened 2025-09-09
Battery values correction - Reload fix

thewhale21:
Update const.py to load switches before sensors to support the KWH_Calculation

Changing the load order has fixed the issue mentioned in #296 and #289 

Is there a list of checks we can do to ensure it hasnt effected any other logic?

### DIFF
--- a/custom_components/stellantis_vehicles/const.py
+++ b/custom_components/stellantis_vehicles/const.py
@@ -88,8 +88,8 @@ PLATFORMS = [
     "button",
     "device_tracker",
     "number",
-    "sensor",
     "switch",
+    "sensor",
     "text",
     "time"
 ]


khenderick:
Updated my env and restarted. I currently have the usual Stellantis issues where the integration can't connect, so let's see what happens when it comes back.

khenderick:
While applying #303 and restarting I noticed my sensors had a brief drop again. So not sure if only changing the order is sufficient. I guess there's a lot of async code running so we apparently don't have real control about what runs when. 

I wonder whether a solution might be to wait for updates until all setting switches are loaded. I'm not sure on whether or how this is possible, though.

thewhale21:
Did the conversion not take place for you on this reload then?

I was trying to get the code to run in a specific order, but it involved changing so many files and i kept seeing bugs so gave up and went with the simplest solution (Albeit maybe not perfect). For me at least it has greatly improved the reliability of the switch.

khenderick:
Unfortunately it indeed only took place after a second update. Maybe the switch can be put on top of the list instead of just above sensor, to increase the chance.

thewhale21:
I guess it wouldnt have any negative effect, The code would still be async, but it would improve the chances of it completing first as it had started first. 

Still not 100% reliable but it would improve the chances.

PR #300, opened 2025-09-09
Unit conversion issue on load

thewhale21:
At present, reloading the intergration will cause double conversion of values that have been changed from thier native unit

For example last trip, if set to miles (Or any other unit) rather than km, reloading the intergration will cause the intergration to take that value and then pass it back to HA, HA then tries to convert it back to what the user has set which can result in the below behavior
image

Another example this time setting the last trip unit to meters
image

The implemented code should stop this by converting back to km. 

This can also apply to sensors other than last_trip such as speed however my car doesnt support these to test.

### DIFF
--- a/custom_components/stellantis_vehicles/base.py
+++ b/custom_components/stellantis_vehicles/base.py
@@ -15,8 +15,9 @@ from homeassistant.components.text import TextEntity
 from homeassistant.components.time import TimeEntity
 from homeassistant.core import callback, HomeAssistant
 from homeassistant.helpers.restore_state import RestoreEntity
-from homeassistant.const import ( STATE_UNAVAILABLE, STATE_UNKNOWN, STATE_ON, STATE_OFF )
+from homeassistant.const import ( STATE_UNAVAILABLE, STATE_UNKNOWN, STATE_ON, STATE_OFF, UnitOfLength, UnitOfSpeed)
 from homeassistant.exceptions import ConfigEntryAuthFailed
+from homeassistant.util.unit_conversion import DistanceConverter, SpeedConverter
 
 from .utils import ( time_from_pt_string, get_datetime, date_from_pt_string, datetime_from_isoformat, time_from_string )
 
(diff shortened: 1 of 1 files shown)


andreadegiovine:
https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/releases/tag/2025.9.6-beta.1

thewhale21:
#289

thewhale21:
> https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/releases/tag/2025.9.6-beta.1

Thank you :)

My stellantis is limiting my refreshes so will get to test this more tomorrow.

Edit:
Hopefully i didnt break anything!

thewhale21:
I’ve switched the restore path to use async_get_last_sensor_data() and assign native_value 

If no native restore data is available, I fall back to async_get_last_state().

Do you think the below would be better?


class StellantisRestoreSensor(StellantisBaseEntity, RestoreSensor):
    async def async_added_to_hass(self):
        """Restore entity data on system restart."""
        await super().async_added_to_hass()

        # Preferred path: restore native sensor data (avoids unit conversion issues)
        sensor_data = await self.async_get_last_sensor_data()
        restored_state = None
        value = None

        if sensor_data is not None and sensor_data.native_value is not None:
            value = sensor_data.native_value
            # If a native unit was stored, ensure our entity reflects it
            if getattr(sensor_data, "native_unit_of_measurement", None):
                self._attr_native_unit_of_measurement = sensor_data.native_unit_of_measurement
        else:
            # Fallback for older stored states (or first run)
            restored_state = await self.async_get_last_state()
            if restored_state and restored_state.state not in [STATE_UNAVAILABLE, STATE_UNKNOWN]:
                value = restored_state.state
                if self._key == "battery_charging_end":
                    value = datetime.fromisoformat(value)

        if value is not None:
            self._attr_native_value = value
            self._coordinator._sensors[self._key] = value

        # Restore attributes from the last state record (if present)
        if restored_state is None:
            restored_state = await self.async_get_last_state()
        if restored_state:
            for key, attr_val in restored_state.attributes.items():
                self._attr_extra_state_attributes[key] = attr_val

        self.coordinator_update()

    def coordinator_update(self):
        """Coordinator update."""
        return True

PR #303, opened 2025-09-10
Avoids manual conversion when restoring sensors

thewhale21:
possible alternative to #300 to avoid manual conversion.


khenderick:
I think this looks more in line with the docs. I'll update my environment and see if I get any issues during restart(s).

khenderick:
This seems to work fine if I change my total trip sensor to miles, and then reload the integration :+1:.

PR #335, opened 2025-10-30
Update updated_at_map keys due to Stellantis/PSA API change

### LINKED ISSUE #331: updated_at attributes are frozen since Stellantis/PSA API change
### Mark

Opel

### Model

all (Corsa-e in my case)

### Year

2020

### Engine

Electric

### Remote service

E-remote control

### HomeAssistant version

2025.10.4

### Integration version

2025.10.3

### What happened?

A while ago, Stellantis/PSA changed their API: Before the change, there was a "createdAt" field in the energies object which contained the timestamp when the data was last obtained from the car. Since the API change, this value does not change anymore. Instead, there is now (a more appropriately named) new "updatedAt" field that contains the expected timestamp.

Currently, this seems to be not reflected in the HA integration, leading to frozen updated_at attributes of several sensors. I guess `["energies", {"type":"Electric"}, "createdAt"]` needs to be replaced by ["energies", {"type":"Electric"}, "updatedAt"] throughout const.py. In [https://github.com/flobz/psa_car_controller/pull/1096](https://github.com/flobz/psa_car_controller/pull/1096) this was handled for psa_car_controller for reference.

### Log output

```shell

```

miradarya:
Fixes #331.

Disclaimer: I don't own a fossil car anymore, so I was only able to verify this works correclty for the Electric type energy metrics but not Fuel type energy metrics. I am pretty confident though that it also works for those.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,4 @@
 __pycache__
+.idea
+.vscode
+data.json
--- a/custom_components/stellantis_vehicles/const.py
+++ b/custom_components/stellantis_vehicles/const.py
@@ -29,9 +29,12 @@ CAR_API_GET_VEHICLE_STATUS_URL = CAR_API_VEHICLES_URL + "/{#vehicle_id#}/status"
 CAR_API_GET_VEHICLE_TRIPS_URL = CAR_API_VEHICLES_URL + "/{#vehicle_id#}/trips"
 
 MQTT_SERVER = "mwa.mpsa.com"
+MQTT_PORT = 8885
+MQTT_KEEP_ALIVE_S = 120
 MQTT_RESP_TOPIC = "psa/RemoteServices/to/cid/"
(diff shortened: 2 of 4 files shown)


andreadegiovine:
Hi @miradarya and thank for your support.
As described on official docs (https://developer.groupe-psa.io/webapi/b2c/api-reference/references/#tag/Vehicles/operation/getVehicleStatus) the "energy" node is listed as deprecated.
I know the data in that node is correct, but do we really want to rely on an entity that is marked as deprecated?
Perhaps it's better to rely on HomeAssistant to realistically track the last update date, what do you think?

miradarya:
Oh, sorry, I didn't realize it is depricated. In that case, we shouldn't use it.

I just thought it was a good way to distinguish between when HomeAssistant last updated the data and when the Stellantis/PSA server actually got the data from the car.

PR #336, opened 2025-11-03
Add async refresh after wakeup command in button.py

### LINKED ISSUE #319: Refresh data immediately after Wakeup Complete
### Mark

Peugeot

### Model

e-208

### Year

2025

### Engine

Electric

### Remote service

E-remote control

### HomeAssistant version

2025.8.3

### Integration version

2025.10.1-beta.1

### What happened?

It would be great to trigger `_async_update_data` automatically after a Wakeup Command completes. Would also make the charge limit function more accurate, since it doesn't wait for another 1-2 minutes before the integration sees the new SOC value.

### Log output

```shell

```

miradarya:
Fixes #319 .

There remains one problem, however: All entities, including the wakeup button, are set to unavailable when the engine is on. This prevents forcing an update while the car is driving until the next, regular refresh interval. Which is unfortunate, because I am trying to use the integration for some control of my wallbox. When I am returning from a trip and plug in the car to the wallbox, it is not possible to refresh the latest car state because the integration has not yet realized the engine is not on anymore.

Therefore, I'd suggest to make the wakeup button available even if the engine is on. psa_car_controller for example always allows forced updates (including wakeup) without any issue. Any thoughts on this?

### DIFF
--- a/custom_components/stellantis_vehicles/button.py
+++ b/custom_components/stellantis_vehicles/button.py
@@ -101,6 +101,7 @@ async def async_setup_entry(hass:HomeAssistant, entry, async_add_entities) -> No
 class StellantisWakeUpButton(StellantisBaseButton):
     async def async_press(self):
         await self._coordinator.send_wakeup_command(self.name)
+        await self._coordinator.async_refresh()
 
 class StellantisDoorButton(StellantisBaseActionButton):
     async def async_press(self):


andreadegiovine:
Refresh only if command is in “complete” status

andreadegiovine:
@miradarya try this https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/releases/tag/2025.11.1-beta.2

miradarya:
Thanks, will do. I was unsure where best to put the refresh.

PR #355, opened 2025-11-19
fix: set rate limit duration for send_wakeup_command to 1200

tamcore:
Assuming it's indeed supposed to be 20 minutes (as the comment and the translations indicate), this should be `1200` seconds, instead of just `120` :)


andreadegiovine:
Can you make pull request from/to develop?

tamcore:
Coming in via #356 :)

PR #450, opened 2026-03-13
Feature implementation - track charge cost and real-world consumption

GabrielRogd:
## Summary

This PR adds charge cost tracking and a real-world average consumption sensor for charge-capable vehicles.

New entities:
- `number.kwh_cost` to set the electricity price per vehicle
- `sensor.charge_cost` to show the cost of the latest completed charge
- `sensor.actual_average_consumption` to show actual average consumption in `kWh/100km`

`sensor.last_charge` now also stores mileage, distance since last charge, charge cost, and actual average consumption.

## Why

This makes it easier to track what a charge really costs and how efficient the car is between charging sessions.

For example, on my Corsa-e 2020, I wanted to see the real consumption based on the energy added during charging and the distance driven since the previous full charge session, instead of only relying on the trip value reported by the app.

## Notes

- Charge cost stays unavailable until `number.kwh_cost` is set above `0`
- Actual average consumption becomes available after enough charge history exists, typically after the second completed charge recorded by this version
- Battery value correction is also applied to the derived charge metrics so the numbers stay consistent

## Validation

- `python -m compileall custom_components/stellantis_vehicles`

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@
 - [OAuth2 Code](#oauth2-code)
 - [Commands](#commands)
 - [Battery capacity / residual sensors](#battery-capacity--residual-sensors)
+- [Charge cost / actual consumption](#charge-cost--actual-consumption)
 - [Card: Stellantis Vehicles](#card-stellantis-vehicles)
 - [Global preferences](#global-preferences)
 - [Errors](#errors)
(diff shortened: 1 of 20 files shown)


Jordan87:
Which values are taken to calculate those values?
My PlugInHybrid (308) is not providing information about Battery residual or Battery capacity.

GabrielRogd:
Sorry for my late answer. This implementation was made using an e-remote control capable Opel Corsa-e, from which the residual/capacity values can be fetched. The charge costs are calculated based on average consumption (shown after every trip, and then the average is calculated during the current charge cycle), and total kWh consumption before the car is put into the wall.

PR #476, opened 2026-05-14
Fix Stellantis API bug for autonomy and battery values

Destroyer061090:
Add a fix for Stellantis API bug related to autonomy and battery values.


andreadegiovine:
Thanks for your support, your fix was committed under "get_value" method like other data manipulations

i close thid pr

Issue #487, opened 2026-06-09
Don't update sensor states when value is unchanged


jostsalathe:
### Mark

Opel

### Model

Corsa-e

### Year

2019

### Engine

Electric

### Remote service

Connect Plus

### HomeAssistant version

2026.6.1

### Integration version

2026.5.1

### What happened?

I use derivative helpers to calculate
- vehicle speed from the odometer value and
- battery power from battery residual.

The derivatives usually produce absurd spikes if I don't configure them to use a 5 minute average. The average is undesirable because it makes the derivative lag behind. This makes it effectively useless for short trips like daily commute that takes 15 minutes or so.

I believe the cause of the spikes are updates of the HA sensor states where the value from the Stellantis API has not actually changed.
I made a little demonstration of what I think is happening in a spreadsheet:

<img width="1272" height="980" alt="Image" src="https://github.com/user-attachments/assets/ea1d0826-81eb-4ed1-a73d-c592c7e78e10" />

The upper example shows what would currently happen. The marked data entry has no value change but of course it has a time stamp change. This might happen when the Stellantis API has not yet received updates from the vehicle over consecutive requests from HA.
The API then reports the old values and this App updates the sensor state which gives the old value a new time stamp.

The lower graph shows what would happen if this App only updates the sensor state when the value actually changes.

I think this has also be considered on App reloads (e.g. after an App update or a HA restart)

I am completely open to discussions on this topic. Maybe there is a better way to fix the issue. I don't actually know how other apps handle those cases.

I have just added a derivative of my odometer sensor for testing with the average option disabled so I can gather some actual examples. But this will take some time and driving around.

### Log output

```shell

```

sjonez:
I think #492 might fix this for you as a side effect, as part of rejecting stale data we also ignore data updates with the same timestamp. Worth checking out once its released

Edit: Just noticed the PR is actually yours! So maybe its not a side effect after all 😆

jostsalathe:
#492 will not completely fix the issue, because the createdAt property that is checked there changes if any sensor value is updated (or maybe even if a new set of data is available without the actual data being changed).

This issue can only be fixed by checking for value changes on each individual entity, I think.

odegaard12:
After validating #492 together with #491 over additional real-world usage, I think the stale payload handling works correctly.

However, I still believe this issue is independent.

[...]

PR #491, opened 2026-06-16
fix: pass config entry explicitly to coordinator

odegaard12:
## Summary

Pass the integration config entry explicitly to
`DataUpdateCoordinator`.

This removes reliance on the implicit config-entry context and prepares
the coordinator wiring for the Home Assistant 2026.8 behavior change.

## Changes

- Add `config_entry` to `StellantisVehicleCoordinator`.
- Pass it to `DataUpdateCoordinator`.
- Pass the integration entry from `StellantisVehicles`.

## Validation

- [x] Python compilation
- [x] Static AST validation
- [x] Home Assistant 2026.6.3 startup
- [x] Existing configuration loads
- [x] Sensors continue updating
- [x] Home Assistant configuration check
- [x] No Stellantis errors in initial logs
- [ ] Seven-day stability period completed
- [ ] New installation tested
- [ ] Reconfiguration tested

Remote commands are disabled and unavailable for the test vehicle, so
no remote command was sent during validation.

PR #492, opened 2026-06-16
added check for old 'createdAt' time stamps to _async_update_data()

### LINKED ISSUE #488: Car data seems to contain stale data points
### Mark

Peugeot

### Model

e-208

### Year

2020

### Engine

Electric

### Remote service

E-remote control

### HomeAssistant version

2026.6.2

### Integration version

2026.5.1

### What happened?

It seems data received from Peugeot contains stale data points inbetween current ones. This means that the battery percentage in HA and ABRP constantly flips between a current value and an old one. As can be seen in the screenshot. This has happened for the past two days now in the afternoon.

This might be a stellantis issue, but I'm wondering if others are experiencing this too and maybe if something can be done about this. 

Image

### Log output

```shell

```

jostsalathe:
This is necessary because the Stellantis API sometimes responds with data sets that are older than previous ones, which may result for example in decreasing odometer values.

Should fix #488

### DIFF
--- a/custom_components/stellantis_vehicles/base.py
+++ b/custom_components/stellantis_vehicles/base.py
@@ -69,7 +69,7 @@ class StellantisVehicleCoordinator(DataUpdateCoordinator):
                         accepted_new_data = True
                     elif new_data_time < old_data_time:
                         # If the received data set is actually older, log a warning
-                        _LOGGER.warning(f"Discarded stale data set - new 'createdAt' too old: self._data: {old_data_time}, new_data: {new_data_time})")
+                        _LOGGER.warning(f"Discarded stale data set - new 'createdAt' too old: self._data: {old_data_time}, new_data: {new_data_time}")
                 else:
                     self._data = new_data
                     accepted_new_data = True
--- a/custom_components/stellantis_vehicles/base.py
+++ b/custom_components/stellantis_vehicles/base.py
@@ -59,20 +59,19 @@ class StellantisVehicleCoordinator(DataUpdateCoordinator):
         try:
             # Vehicle status
             new_data = await self._stellantis.get_vehicle_status(self._vehicle)
-            if "createdAt" in new_data:
-                if "createdAt" in self._data:
-                    old_data_time = datetime.fromisoformat(self._data["createdAt"])
-                    new_data_time = datetime.fromisoformat(new_data["createdAt"])
-                    if new_data_time > old_data_time:
-                        # Only accept new data sets that are actually newer than the last one
-                        self._data = new_data
-                        accepted_new_data = True
-                    elif new_data_time < old_data_time:
-                        # If the received data set is actually older, log a warning
-                        _LOGGER.warning(f"Discarded stale data set - new 'createdAt' too old: self._data: {old_data_time}, new_data: {new_data_time}")
-                else:
+            if "createdAt" in new_data and "createdAt" in self._data:
+                old_data_time = datetime.fromisoformat(self._data["createdAt"])
+                new_data_time = datetime.fromisoformat(new_data["createdAt"])
+                if new_data_time > old_data_time:
+                    # Only accept new data sets that are actually newer than the last one
                     self._data = new_data
                     accepted_new_data = True
+                elif new_data_time < old_data_time:
+                    # If the received data set is actually older, log a warning
+                    _LOGGER.warning(f"Discarded stale data set - new 'createdAt' too old: self._data: {old_data_time}, new_data: {new_data_time}")
+            else:
+                self._data = new_data
+                accepted_new_data = True
         except ConfigEntryAuthFailed:
             _LOGGER.debug("---------- END _async_update_data")
             raise
--- a/custom_components/stellantis_vehicles/base.py
+++ b/custom_components/stellantis_vehicles/base.py
@@ -55,6 +55,7 @@ class StellantisVehicleCoordinator(DataUpdateCoordinator):
         """ Update vehicle data from Stellantis. """
         _LOGGER.debug("---------- START _async_update_data")
         _LOGGER.debug(self._config)
+        accepted_new_data = False
         try:
             # Vehicle status
             new_data = await self._stellantis.get_vehicle_status(self._vehicle)
@@ -65,6 +66,7 @@ class StellantisVehicleCoordinator(DataUpdateCoordinator):
                     if new_data_time > old_data_time:
                         # Only accept new data sets that are actually newer than the last one
                         self._data = new_data
+                        accepted_new_data = True
                     elif new_data_time < old_data_time:
                         # If the received data set is actually older, log a warning
                         _LOGGER.warning(f"Discarded stale data set - new 'createdAt' too old: self._data: {old_data_time}, new_data: {new_data_time})")
@@ -75,7 +77,8 @@ class StellantisVehicleCoordinator(DataUpdateCoordinator):
             raise
         except Exception:
             pass
-        await self.after_async_update_data()
+        if accepted_new_data:
+            await self.after_async_update_data()
         _LOGGER.debug("---------- END _async_update_data")
 
     def get_translation(self, path, default = None):
--- a/custom_components/stellantis_vehicles/base.py
+++ b/custom_components/stellantis_vehicles/base.py
@@ -63,11 +63,11 @@ class StellantisVehicleCoordinator(DataUpdateCoordinator):
                     old_data_time = datetime.fromisoformat(self._data["createdAt"])
                     new_data_time = datetime.fromisoformat(new_data["createdAt"])
                     if new_data_time > old_data_time:
+                        # Only accept new data sets that are actually newer than the last one
                         self._data = new_data
                     elif new_data_time < old_data_time:
-                        _LOGGER.warning("Discarded stale data set - received 'createdAt' too old:")
-                        _LOGGER.warning(f"  self._data['createdAt']: {old_data_time}")
-                        _LOGGER.warning(f"    new_data['createdAt']: {new_data_time})")
+                        # If the received data set is actually older, log a warning
+                        _LOGGER.warning(f"Discarded stale data set - new 'createdAt' too old: self._data: {old_data_time}, new_data: {new_data_time})")
                 else:
                     self._data = new_data
         except ConfigEntryAuthFailed:
--- a/custom_components/stellantis_vehicles/base.py
+++ b/custom_components/stellantis_vehicles/base.py
@@ -70,6 +70,7 @@ class StellantisVehicleCoordinator(DataUpdateCoordinator):
                     elif new_data_time < old_data_time:
                         # If the received data set is actually older, log a warning
                         _LOGGER.debug(f"Discarded stale data set - new 'createdAt' too old: self._data: {old_data_time}, new_data: {new_data_time}")
+                    # Discard data sets that have the same 'createdAt' value as the last accepted data set (not new but also not old)
                 else:
                     self._data = new_data
                     accepted_new_data = True
--- a/custom_components/stellantis_vehicles/base.py
+++ b/custom_components/stellantis_vehicles/base.py
@@ -68,7 +68,7 @@ class StellantisVehicleCoordinator(DataUpdateCoordinator):
                     accepted_new_data = True
                 elif new_data_time < old_data_time:
                     # If the received data set is actually older, log a warning
-                    _LOGGER.warning(f"Discarded stale data set - new 'createdAt' too old: self._data: {old_data_time}, new_data: {new_data_time}")
+                    _LOGGER.debug(f"Discarded stale data set - new 'createdAt' too old: self._data: {old_data_time}, new_data: {new_data_time}")
             else:
                 self._data = new_data
                 accepted_new_data = True
--- a/custom_components/stellantis_vehicles/base.py
+++ b/custom_components/stellantis_vehicles/base.py
@@ -72,6 +72,7 @@ class StellantisVehicleCoordinator(DataUpdateCoordinator):
                         _LOGGER.warning(f"Discarded stale data set - new 'createdAt' too old: self._data: {old_data_time}, new_data: {new_data_time})")
                 else:
                     self._data = new_data
+                    accepted_new_data = True
         except ConfigEntryAuthFailed:
             _LOGGER.debug("---------- END _async_update_data")
             raise
--- a/custom_components/stellantis_vehicles/base.py
+++ b/custom_components/stellantis_vehicles/base.py
@@ -57,7 +57,19 @@ class StellantisVehicleCoordinator(DataUpdateCoordinator):
         _LOGGER.debug(self._config)
         try:
             # Vehicle status
-            self._data = await self._stellantis.get_vehicle_status(self._vehicle)
+            new_data = await self._stellantis.get_vehicle_status(self._vehicle)
+            if "createdAt" in new_data:
+                if "createdAt" in self._data:
+                    old_data_time = datetime.fromisoformat(self._data["createdAt"])
+                    new_data_time = datetime.fromisoformat(new_data["createdAt"])
+                    if new_data_time > old_data_time:
+                        self._data = new_data
+                    elif new_data_time < old_data_time:
+                        _LOGGER.warning("Discarded stale data set - received 'createdAt' too old:")
+                        _LOGGER.warning(f"  self._data['createdAt']: {old_data_time}")
+                        _LOGGER.warning(f"    new_data['createdAt']: {new_data_time})")
+                else:
+                    self._data = new_data
         except ConfigEntryAuthFailed:
             _LOGGER.debug("---------- END _async_update_data")
             raise
--- a/custom_components/stellantis_vehicles/base.py
+++ b/custom_components/stellantis_vehicles/base.py
@@ -59,19 +59,25 @@ class StellantisVehicleCoordinator(DataUpdateCoordinator):
         try:
             # Vehicle status
             new_data = await self._stellantis.get_vehicle_status(self._vehicle)
-            if "createdAt" in new_data and "createdAt" in self._data:
-                old_data_time = datetime.fromisoformat(self._data["createdAt"])
-                new_data_time = datetime.fromisoformat(new_data["createdAt"])
-                if new_data_time > old_data_time:
-                    # Only accept new data sets that are actually newer than the last one
+            try:
+                if "createdAt" in new_data and "createdAt" in self._data:
+                    old_data_time = datetime.fromisoformat(self._data["createdAt"])
+                    new_data_time = datetime.fromisoformat(new_data["createdAt"])
+                    if new_data_time > old_data_time:
+                        # Only accept new data sets that are actually newer than the last one
+                        self._data = new_data
+                        accepted_new_data = True
+                    elif new_data_time < old_data_time:
+                        # If the received data set is actually older, log a warning
+                        _LOGGER.debug(f"Discarded stale data set - new 'createdAt' too old: self._data: {old_data_time}, new_data: {new_data_time}")
+                else:
                     self._data = new_data
                     accepted_new_data = True
-                elif new_data_time < old_data_time:
-                    # If the received data set is actually older, log a warning
-                    _LOGGER.debug(f"Discarded stale data set - new 'createdAt' too old: self._data: {old_data_time}, new_data: {new_data_time}")
-            else:
+                    _LOGGER.debug("Accepted data set with missing 'createdAt'")
+            except ValueError:
                 self._data = new_data
                 accepted_new_data = True
+                _LOGGER.debug("Accepted data set with corrupt 'createdAt'")
         except ConfigEntryAuthFailed:
             _LOGGER.debug("---------- END _async_update_data")
             raise


odegaard12:
I tested the actual code from this PR in isolation using synthetic vehicle-status payloads.

Results:

- First payload is accepted.
- A payload with a newer `createdAt` is accepted.
- A payload with an older `createdAt` is correctly discarded.
- Python compilation passes.
- The main stale-data protection behaves as intended.

I also observed a few edge cases worth considering:

1. A payload without a top-level `createdAt` is never accepted when previous data already exists. If the API temporarily omits that field, telemetry could remain frozen indefinitely.
2. A payload with the same `createdAt` is also ignored. This may be intentional, but it would be useful to document the expected behavior.
3. Each discarded payload currently emits three warning lines. During the current Stellantis incident this could generate a large amount of log noise. A single parameterized warning, or a debug/rate-limited message, may be preferable.
4. `after_async_update_data()` still runs after rejecting a stale payload, using the previously accepted data. This may be fine, but it is worth confirming that downstream actions should still execute for a rejected update.

The core fix is effective against the stale payload sequence I reproduced on a 2022 Citroën ë-C4.

sjonez:
ignoring duplicate timestamps might actually be a good thing and could fix #487 perhaps?

odegaard12:
Yes, potentially.

The synthetic test confirmed that a payload with the same `createdAt` is ignored, so it may prevent identical API payloads from replacing the previously accepted data.

However, I would not consider #487 fully resolved yet. In the current implementation, `after_async_update_data()` still runs after an equal or stale payload is rejected, and coordinator listeners may still be notified with the previously accepted data.

So this PR may reduce duplicate updates, but the remaining Recorder/state-write behaviour should be measured after applying it before closing #487.

I can help test that specific sequence if the PR is updated or a test branch is made available.

jostsalathe:
Thank you for your suggestions, @odegaard12. I addressed the following of your identified edge cases:
1. I think data sets that omit the top level `createdAt` would probably be corrupt enough that they should be ignored, anyway.
    - I am thinking about adding backup checks for `createdAt` entries deeper down in the data set structure. Do you have suggestions on how to implement that?
2. That payloads with the same `createdAt` value are ignored is intentional and I added comments about it to document the intention.
3. I condensed the three warning messages into one so that the logger is less polluted when an incident occurs
4. `after_async_update_data()` now gets executed only if a new data set was actually accepted.

odegaard12:
Thanks for addressing those points. The changes for equal timestamps, the condensed warning and only running after_async_update_data() after accepting new data all make sense.

I noticed one small first-update case in the latest version.

Since self._data starts as an empty dictionary, the first payload enters the initial-accept branch and stores new_data, but accepted_new_data remains False. That means after_async_update_data() is not executed for that first accepted payload. If the next API response has the same createdAt, it will also be ignored.

I think the first-accept branch should also set accepted_new_data = True.

For deeper timestamps, I would avoid recursively selecting any createdAt, because different vehicle sections may legitimately update independently. A safer follow-up could compare a small explicit set of timestamps against the same paths in the previously accepted payload, for example odometer.createdAt, lastPosition.properties.createdAt, and the relevant energy timestamp.

I would probably keep the top-level check as the main safeguard in this PR and treat the nested checks as a separate follow-up once the relevant paths have been confirmed across several vehicle payloads.

I can retest the synthetic cases after the first-update branch is adjusted.

jostsalathe:
Good observation, thank you @odegaard12.
I added the `accepted_new_data = True` statement also to the first data branch.

Another test on your synthetic cases might be useful now, yes.

odegaard12:
I retested the updated PR after the first-accepted-payload fix.

Validation performed:

- the initial accept branch now sets accepted_new_data = True;
- newer payloads still set accepted_new_data = True;
- after_async_update_data() is gated by accepted_new_data;
- the stale-data warning is condensed to one stale-data warning;
- synthetic sequence covering first, newer, older, equal-timestamp and missing-createdAt payloads.

Result: OK

The first-update case I mentioned is fixed now, and the synthetic stale-data sequence behaves as expected.

I think the top-level createdAt guard is now good as the main safeguard for this PR. Deeper per-section timestamp checks can probably be treated as a separate follow-up if needed.

andreadegiovine:
@jostsalathe 

Great job!

My thoughts:
We're relying on the value of createdAt (among all the open issues, I've noticed that it's not always present in the various nodes), and based on that data, we decide whether to update the entities. 

But what if Stellantis were to encounter a bug and that value remained unchanged?

There are issues related to updating entities when Stellantis doesn't provide values ​​(entities go to Unknown even if a state has been manually set). 

What do you think about trying to fix this bug in a new PR using a wrapper method, and preventing incorrect value updates on the specific keys in the method? 

For example: If Stellantis doesn't provide the value, don't update the entity.

If Stellantis provides a value lower than the current one and the key is mileage, don't update the entity.

odegaard12:
That makes sense to me.

I agree that this should probably be handled separately from this PR. The current PR is a good top-level stale-payload safeguard, while the missing/invalid value handling sounds better as a follow-up wrapper around specific keys.

For example:
- missing value: keep the previous valid state instead of updating to unknown;
- mileage lower than the current accepted value: ignore that specific value;
- other keys could be handled conservatively only when the invalid pattern is well understood.

I would keep #492 focused on the createdAt guard and test the per-key wrapper separately.

jostsalathe:
Thanks, @andreadegiovine!

Okay, so you mean we would change `_async_update_data()` so that it
- still rejects whole data sets with a top level `createdAt` that is too old but
- accepts data sets that have no top level `createdAt`
to make it more robust against API bugs where the top level `createdAt` might be missing.

As extra layer of protection against stale data contained in data sets that have no top level `createdAt`, we would - in a separate PR - implement per entity checks against local (associated directly with that entity) `createdAt` values that are too old (where they are available) as well as entity values that are implausible (where it is possible - e.g. decreasing odometer)?

In #488 always the _complete_ data set seems to be out of date if it happens. So I assumed that we would want to reject the whole packet if a single implausible data point is detected.

Sphinxapfel7:
I can confirm the issue this PR addresses on my Opel Astra L Sportstourer Electric. While parked with the charging cable connected, HA constantly toggles battery_plugged and preconditioning states due to stale API responses.

Thank you for working on this fix – looking forward to seeing it merged!

odegaard12:
## Additional validation

I tested this implementation together with #491 on my Home Assistant OS 2026.6.3 installation.

During additional validation I observed:

- clean startup;
- successful integration reloads;
- correct coordinator initialization;
- successful Home Assistant Core restart;
- multiple polling cycles during normal vehicle usage;
- no startup regressions;
- no unexpected coordinator failures.

The stale payload protection behaves as expected and I have not observed the previous stale data flip-flop behaviour again during normal usage.

The additional handling of payloads without a top-level `createdAt` also improves robustness against incomplete API responses.

custom_components/stellantis_vehicles/base.py
```diff
@@ -55,15 +55,30 @@ async def _async_update_data(self):
         """ Update vehicle data from Stellantis. """
         _LOGGER.debug("---------- START _async_update_data")
         _LOGGER.debug(self._config)
+        accepted_new_data = False
```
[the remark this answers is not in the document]
andreadegiovine: And rename it like “new_data_received” 😀

phryneas:
I did some testing on this, too (actually I have my own solution and just wanted to open a PR, but I guess I can scrap that ^^).
In my logfiles, I found two instances of log data where `createdAt` and `updatedAt` deviated from each other, and `updatedAt` was hours in the past from `createdAt`.

So I'd assume hat `updatedAt` is the date that the data was retrieved, and `createdAt` as a date it was probably first fully accumulated for sending. So it might be better to use `updatedAt` than `createdAt`.

Data in question:
```json
[
    {
        "createdAt": "2026-06-23T19:30:15Z",
        "updatedAt": "2026-06-23T17:43:51Z",
        "adas": {
            "createdAt": "2026-06-23T17:43:51Z",
            "parkAssist": {
                "front": [
                    "OutOfService"
                ],
                "rear": [
                    "OutOfService"
                ]
            },
            "accr": "Off",
            "lvv": false,
            "rvv": "Off",
            "aebs": "Off",
            "afil": "Off",
            "artiv": "NotSelected",
            "bsm": "Inactive",
            "esp": false,
            "abs": false,
            "fse": true,
            "sli": 30,
            "rlka": "NotSelected",
            "llka": "NotSelected",
            "rgi": "None"
        },
        "lastPosition": {
            "type": "Feature",
            "geometry": {
                "type": "Point",
                "coordinates": [
                    12.34567,
                    12.34567,
                    123.4
                ]
            },
            "properties": {
                "createdAt": "2026-06-23T17:41:14Z",
                "heading": 228,
                "type": "Acquire",
                "signalQuality": 9,
                "fixStatus": "3D"
            }
        },
        "engines": [
            {
                "type": "Electric",
                "speed": 0.0,
                "gmpStatus": "Stopped",
                "createdAt": "2026-06-23T17:43:51Z"
            },
            {
                "type": "Electric",
                "gmpStatus": "Stopped",
                "createdAt": "2026-06-23T17:43:51Z"
            }
        ],
        "powertrain": {
            "createdAt": "2026-06-23T17:43:51Z",
            "status": "Running"
        },
        "ignition": {
            "createdAt": "2026-06-23T17:43:31Z",
            "type": "Stop"
        },
        "battery": {
            "voltage": 84.5,
            "createdAt": "2026-06-23T17:43:31Z"
        },
        "lightingSystem": {
            "turn": [
                {
                    "direction": "Right",
                    "position": "Front",
                    "status": "Off"
                },
                {
                    "direction": "Right",
                    "position": "Rear",
                    "status": "Off"
                },
                {
                    "direction": "Left",
                    "position": "Front",
                    "status": "Off"
                },
                {
                    "direction": "Left",
                    "position": "Rear",
                    "status": "Off"
                }
            ],
            "fog": [
                {
                    "direction": "Left",
                    "position": "Front",
                    "status": "Off"
                },
                {
                    "direction": "Right",
                    "position": "Front",
                    "status": "Off"
                },
                {
                    "direction": "Left",
                    "position": "Rear",
                    "status": "Off"
                },
                {
                    "direction": "Right",
                    "position": "Rear",
                    "status": "Off"
                }
            ],
            "createdAt": "2026-06-23T17:43:51Z"
        },
        "service": {
            "createdAt": "2026-06-23T17:43:51Z",
            "type": "Electric"
        },
        "safety": {
            "createdAt": "2026-06-23T17:43:51Z",
            "beltStatus": [
                {
                    "belt": "Omission",
                    "id": "Driver"
                },
                {
                    "belt": "Omission",
                    "id": "Passenger"
                }
            ],
            "autoECallTriggering": "NotDetected"
        },
        "environment": {
            "luminosity": {
                "createdAt": "2026-06-23T17:43:31Z",
                "day": true
            },
            "air": {
                "createdAt": "2026-06-23T17:43:31Z",
                "temp": 33.0
            }
        },
        "odometer": {
            "createdAt": "2026-06-23T17:43:31Z",
            "mileage": 123456.7
        },
        "kinetic": {
            "createdAt": "2026-06-23T17:43:51Z",
            "acceleration": 0.0,
            "lateralAcceleration": 0.15,
            "speed": 0.0,
            "moving": false
        },
        "_links": {
            "self": {
                "href": "https://api.groupe-psa.com/connectedcar/v4/user/vehicles/########/status?profile=endUser"
            },
            "vehicle": {
                "href": "https://api.groupe-psa.com/connectedcar/v4/user/vehicles/########"
            }
        },
        "preconditioning": {
            "airConditioning": {
                "createdAt": "2026-06-23T17:43:32Z",
                "updatedAt": "2026-06-23T17:43:32Z",
                "status": "Disabled",
                "programs": [
                    {
                        "enabled": false,
                        "slot": 1,
                        "recurrence": "Daily",
                        "start": "PT6H10M",
                        "occurence": {
                            "day": [
                                "Mon",
                                "Tue",
                                "Wed",
                                "Thu",
                                "Fri"
                            ]
                        }
                    },
                    {
                        "enabled": false,
                        "slot": 2,
                        "recurrence": "Daily",
                        "start": "PT14H",
                        "occurence": {
                            "day": [
                                "Mon",
                                "Tue",
                                "Wed",
                                "Thu",
                                "Fri",
                                "Sat"
                            ]
                        }
                    }
                ]
            }
        },
        "energies": [
            {
                "createdAt": "2026-06-23T17:43:32Z",
                "type": "Electric",
                "subType": "ElectricEnergy",
                "level": 30.0,
                "autonomy": 58,
                "extension": {
                    "electric": {
                        "battery": {
                            "load": {
                                "createdAt": "2026-06-23T17:43:32Z",
                                "capacity": 31712,
                                "residual": 8512
                            },
                            "health": {
                                "createdAt": "2026-06-23T17:43:32Z",
                                "resistance": 89
                            }
                        },
                        "charging": {
                            "plugged": true,
                            "status": "InProgress",
                            "remainingTime": "PT2H50M",
                            "chargingRate": 64,
                            "chargingMode": "Slow",
                            "nextDelayedTime": "PT8H",
                            "schedule": []
                        }
                    }
                }
            }
        ],
        "preconditionning": {
            "airConditioning": {
                "createdAt": "2026-06-23T17:43:32Z",
                "updatedAt": "2026-06-23T17:43:32Z",
                "status": "Disabled",
                "programs": [
                    {
                        "enabled": false,
                        "slot": 1,
                        "recurrence": "Daily",
                        "start": "PT6H10M",
                        "occurence": {
                            "day": [
                                "Mon",
                                "Tue",
                                "Wed",
                                "Thu",
                                "Fri"
                            ]
                        }
                    },
                    {
                        "enabled": false,
                        "slot": 2,
                        "recurrence": "Daily",
                        "start": "PT14H",
                        "occurence": {
                            "day": [
                                "Mon",
                                "Tue",
                                "Wed",
                                "Thu",
                                "Fri",
                                "Sat"
                            ]
                        }
                    }
                ]
            }
        },
        "energy": [
            {
                "createdAt": "2026-06-23T17:43:32Z",
                "updatedAt": "2026-06-23T17:43:32Z",
                "type": "Electric",
                "level": 30.0,
                "autonomy": 58,
                "charging": {
                    "plugged": true,
                    "status": "InProgress",
                    "remainingTime": "PT2H50M",
                    "chargingRate": 64,
                    "chargingMode": "Slow",
                    "nextDelayedTime": "PT8H"
                },
                "battery": {
                    "load": {
                        "capacity": 31712,
                        "residual": 8512
                    },
                    "health": {
                        "resistance": 89
                    }
                }
            }
        ]
    },
    {
        "createdAt": "2026-06-26T06:55:16Z",
        "updatedAt": "2026-06-25T19:56:49Z",
        "adas": {
            "createdAt": "2026-06-25T19:56:49Z",
            "parkAssist": {
                "front": [
                    "OutOfService"
                ],
                "rear": [
                    "OutOfService"
                ]
            },
            "accr": "Off",
            "lvv": false,
            "rvv": "Off",
            "aebs": "Off",
            "afil": "Off",
            "artiv": "NotSelected",
            "bsm": "Inactive",
            "esp": false,
            "abs": false,
            "fse": true,
            "sli": 30,
            "rlka": "NotSelected",
            "llka": "NotSelected",
            "rgi": "None"
        },
        "lastPosition": {
            "type": "Feature",
            "geometry": {
                "type": "Point",
                "coordinates": [
                    12.34567,
                    12.34567,
                    123.4
                ]
            },
            "properties": {
                "createdAt": "2026-06-25T19:54:15Z",
                "heading": 198,
                "type": "Acquire",
                "signalQuality": 9,
                "fixStatus": "3D"
            }
        },
        "engines": [
            {
                "type": "Electric",
                "speed": 0.0,
                "gmpStatus": "Stopped",
                "createdAt": "2026-06-25T19:56:49Z"
            },
            {
                "type": "Electric",
                "gmpStatus": "Stopped",
                "createdAt": "2026-06-25T19:56:49Z"
            }
        ],
        "powertrain": {
            "createdAt": "2026-06-25T19:56:49Z",
            "status": "Stopped"
        },
        "ignition": {
            "createdAt": "2026-06-25T19:56:29Z",
            "type": "Stop"
        },
        "battery": {
            "voltage": 80.5,
            "createdAt": "2026-06-25T19:56:29Z"
        },
        "lightingSystem": {
            "turn": [
                {
                    "direction": "Right",
                    "position": "Front",
                    "status": "Off"
                },
                {
                    "direction": "Right",
                    "position": "Rear",
                    "status": "Off"
                },
                {
                    "direction": "Left",
                    "position": "Front",
                    "status": "Off"
                },
                {
                    "direction": "Left",
                    "position": "Rear",
                    "status": "Off"
                }
            ],
            "fog": [
                {
                    "direction": "Left",
                    "position": "Front",
                    "status": "Off"
                },
                {
                    "direction": "Right",
                    "position": "Front",
                    "status": "Off"
                },
                {
                    "direction": "Left",
                    "position": "Rear",
                    "status": "Off"
                },
                {
                    "direction": "Right",
                    "position": "Rear",
                    "status": "Off"
                }
            ],
            "createdAt": "2026-06-25T19:56:49Z"
        },
        "service": {
            "createdAt": "2026-06-25T19:56:49Z",
            "type": "Electric"
        },
        "safety": {
            "createdAt": "2026-06-25T19:56:49Z",
            "beltStatus": [
                {
                    "belt": "Omission",
                    "id": "Driver"
                },
                {
                    "belt": "Omission",
                    "id": "Passenger"
                }
            ],
            "autoECallTriggering": "NotDetected"
        },
        "environment": {
            "luminosity": {
                "createdAt": "2026-06-25T19:56:29Z",
                "day": false
            },
            "air": {
                "createdAt": "2026-06-25T19:56:29Z",
                "temp": 29.5
            }
        },
        "odometer": {
            "createdAt": "2026-06-25T19:56:29Z",
            "mileage": 123456.7
        },
        "kinetic": {
            "createdAt": "2026-06-25T19:56:49Z",
            "acceleration": 0.0,
            "lateralAcceleration": -0.36,
            "speed": 0.0,
            "moving": false
        },
        "_links": {
            "self": {
                "href": "https://api.groupe-psa.com/connectedcar/v4/user/vehicles/########/status?profile=endUser"
            },
            "vehicle": {
                "href": "https://api.groupe-psa.com/connectedcar/v4/user/vehicles/########"
            }
        },
        "preconditioning": {
            "airConditioning": {
                "createdAt": "2026-06-25T19:56:29Z",
                "updatedAt": "2026-06-25T19:56:29Z",
                "status": "Disabled",
                "programs": [
                    {
                        "enabled": false,
                        "slot": 1,
                        "recurrence": "Daily",
                        "start": "PT6H10M",
                        "occurence": {
                            "day": [
                                "Mon",
                                "Tue",
                                "Wed",
                                "Thu",
                                "Fri"
                            ]
                        }
                    },
                    {
                        "enabled": false,
                        "slot": 2,
                        "recurrence": "Daily",
                        "start": "PT14H",
                        "occurence": {
                            "day": [
                                "Mon",
                                "Tue",
                                "Wed",
                                "Thu",
                                "Fri",
                                "Sat"
                            ]
                        }
                    }
                ]
            }
        },
        "energies": [
            {
                "createdAt": "2026-06-25T19:56:29Z",
                "type": "Electric",
                "subType": "ElectricEnergy",
                "level": 54.0,
                "autonomy": 122,
                "extension": {
                    "electric": {
                        "battery": {
                            "load": {
                                "createdAt": "2026-06-25T19:56:29Z",
                                "capacity": 31712,
                                "residual": 15520
                            },
                            "health": {
                                "createdAt": "2026-06-25T19:56:29Z",
                                "resistance": 89
                            }
                        },
                        "charging": {
                            "plugged": false,
                            "status": "Disconnected",
                            "chargingRate": 0,
                            "chargingMode": "No",
                            "nextDelayedTime": "PT8H",
                            "schedule": []
                        }
                    }
                }
            }
        ],
        "preconditionning": {
            "airConditioning": {
                "createdAt": "2026-06-25T19:56:29Z",
                "updatedAt": "2026-06-25T19:56:29Z",
                "status": "Disabled",
                "programs": [
                    {
                        "enabled": false,
                        "slot": 1,
                        "recurrence": "Daily",
                        "start": "PT6H10M",
                        "occurence": {
                            "day": [
                                "Mon",
                                "Tue",
                                "Wed",
                                "Thu",
                                "Fri"
                            ]
                        }
                    },
                    {
                        "enabled": false,
                        "slot": 2,
                        "recurrence": "Daily",
                        "start": "PT14H",
                        "occurence": {
                            "day": [
                                "Mon",
                                "Tue",
                                "Wed",
                                "Thu",
                                "Fri",
                                "Sat"
                            ]
                        }
                    }
                ]
            }
        },
        "energy": [
            {
                "createdAt": "2026-06-25T19:56:29Z",
                "updatedAt": "2026-06-25T19:56:29Z",
                "type": "Electric",
                "level": 54.0,
                "autonomy": 122,
                "charging": {
                    "plugged": false,
                    "status": "Disconnected",
                    "chargingRate": 0,
                    "chargingMode": "No",
                    "nextDelayedTime": "PT8H"
                },
                "battery": {
                    "load": {
                        "capacity": 31712,
                        "residual": 15520
                    },
                    "health": {
                        "resistance": 89
                    }
                }
            }
        ]
    }
]
```

custom_components/stellantis_vehicles/base.py
```diff
@@ -55,15 +55,30 @@ async def _async_update_data(self):
...
+            if "createdAt" in new_data and "createdAt" in self._data:
+                old_data_time = datetime.fromisoformat(self._data["createdAt"])
+                new_data_time = datetime.fromisoformat(new_data["createdAt"])
+                if new_data_time > old_data_time:
+                    # Only accept new data sets that are actually newer than the last one
+                    self._data = new_data
+                    accepted_new_data = True
+                elif new_data_time < old_data_time:
+                    # If the received data set is actually older, log a warning
+                    _LOGGER.debug(f"Discarded stale data set - new 'createdAt' too old: self._data: {old_data_time}, new_data: {new_data_time}")
```
MoellerDi: I am not sure if this is intentional, but cases where new_data_time == old_data_time are silently ignored.
Maybe this should be like `new_data_time <= old_data_time`?

custom_components/stellantis_vehicles/base.py
```diff
@@ -55,15 +55,30 @@ async def _async_update_data(self):
...
+                self._data = new_data
+                accepted_new_data = True
         except ConfigEntryAuthFailed:
             _LOGGER.debug("---------- END _async_update_data")
             raise
         except Exception:
             pass
-        await self.after_async_update_data()
+        if accepted_new_data:
+            await self.after_async_update_data()
```
MoellerDi, line 88: Instead of this solution, I suggest restructuring the if/elif blocks above and—if `new_data_time <= old_data_time` holds true—incorporating logging and an early return. This could improve the code's readability and maintainability.

custom_components/stellantis_vehicles/base.py
```diff
@@ -55,15 +55,30 @@ async def _async_update_data(self):
...
         _LOGGER.debug("---------- START _async_update_data")
         _LOGGER.debug(self._config)
+        accepted_new_data = False
         try:
             # Vehicle status
-            self._data = await self._stellantis.get_vehicle_status(self._vehicle)
+            new_data = await self._stellantis.get_vehicle_status(self._vehicle)
+            if "createdAt" in new_data and "createdAt" in self._data:
+                old_data_time = datetime.fromisoformat(self._data["createdAt"])
+                new_data_time = datetime.fromisoformat(new_data["createdAt"])
```
MoellerDi: To be absolutely on the safe side, one could check for `ValueError` – though that might be a bit of an overreaction.

MoellerDi (commented):
Good work – I just wanted to briefly share my opinion on it.

BlythMeister:
Is this something we can get merged (even with follow-ups)

Having a new release just ship, as soon as someone updates, the lose this change if they applied manually.

If they apply the file from here, they lose the shipped changes.

custom_components/stellantis_vehicles/base.py
jostsalathe: I applied your suggestion with 5a9df6f3036b09e35914579e8003826686b71a96

custom_components/stellantis_vehicles/base.py
jostsalathe: Yes, this is intentional. I added a comment to explain the intention with 0af1885b61926dd5eed59d6ed9e067079bb25caa

custom_components/stellantis_vehicles/base.py
jostsalathe: I currently don't have the time and thought resources to think this through. If you have a concrete implementation proposal, I would be happy to apply it. Otherwise I would tend to postpone the refactoring to a later PR, because a lot of users wish to see the changes of this PR in a release.

MoellerDi: I made a suggestion in PR #499 a few weeks ago. However, I agree with you: a fix should be implemented soon.
@andreadegiovine what is missing from your perspective?

PR #510, opened 2026-07-18
Support custom self-hosted login-service URL for remote OAuth

chmielowiec:
Adds an optional field to the "remote service" OAuth step so users can point it at their own self-hosted deployment of the worker instead of the shared Render.com instance.

No behavior change for existing users - the field defaults to the current Render.com URL, so anyone who doesn't touch it gets identical behavior to before.


chmielowiec:
not stale

andreadegiovine:
@chmielowiec seem to be ok, good work!

after my holidays it will be merged

PR #515, opened 2026-07-25
fix button entity name in vehicle card yaml: start_charge -> charge_start

va1entin:
Setting this up with MyOpel I noticed that one button entity in the YAML for the vehicle card didn't exist. It seems to be a simple mixup saying `start_charge` instead of `charge_start`.

Thanks for the amazing integration! :heart:


va1entin:
Not stale

PR #521, opened 2026-08-02
Fail fast on worker timeout, accept full URL in manual OAuth code

the-baaron:
## Summary
- Reduces the OAuth worker request timeout from 300s to 60s (`stellantis.py`). The worker (`homeassistant-stellantis-vehicles-worker.onrender.com`) can hang indefinitely instead of returning an error, so a failed login currently means waiting up to 5 minutes before seeing any feedback. 60s still gives a cold-starting free-tier instance time to wake up, but fails fast otherwise using the existing `ComunicationError("Request timeout")` path — no new error handling needed.
- Lets the manual OAuth code field (`config_flow.py`, `async_step_oauth_manual`) accept the whole redirect URL instead of requiring users to hand-extract the `code` query param. The post-login redirect goes to a custom app scheme (e.g. `mymacsdk://oauth2redirect/...`) that desktop/mobile browsers can't open, so in practice users copy the entire broken-link URL from the address bar rather than picking out just the code. A small helper parses `code=` out of it if present; a bare code still works exactly as before.

Both changes are small and isolated — no changes to the auth flow logic itself, just timeout tuning and input parsing.

## Test plan
- [x] Verified both files parse (`ast.parse`) with no syntax errors
- [x] Applied both changes locally against a live HA instance running this integration (version 2026.7.2); HA restarts cleanly and the integration loads without errors
- [ ] Would appreciate a maintainer/other user confirming the manual-mode paste-URL flow end-to-end with a real login, since I hit the worker being unresponsive rather than the manual path itself

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/custom_components/stellantis_vehicles/config_flow.py
+++ b/custom_components/stellantis_vehicles/config_flow.py
@@ -1,6 +1,7 @@
 import logging
 import voluptuous as vol
 from datetime import timedelta
+from urllib.parse import urlparse, parse_qs
 from uuid import uuid4
 
 from homeassistant.config_entries import ( ConfigFlow, SOURCE_REAUTH, SOURCE_RECONFIGURE )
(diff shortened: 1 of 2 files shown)


the-baaron:
Closing this in favor of a different approach — after more digging, the timeout/paste tweaks here are just band-aids on the underlying issue (the external worker's login automation is fundamentally fragile). Working on a PR that implements the OAuth flow directly instead of depending on the worker.

Issue #527, opened 2026-08-14
Stellantis disables connectivity after 7 days without a 15-minute trip — worth documenting, and probably detectable


kZou13:
### Mark

Peugeot

### Model

3008

### Year

2020

### Engine

Hybrid

### Remote service

E-remote control

### HomeAssistant version

2026.8.1

### Integration version

2026.7.2

### What happened?

Hi,

This isn't a bug in the integration — it's a Stellantis behaviour that made me think the integration was broken for two days. I think it deserves a line in the README, and maybe more.

What happened

All my remote commands were failing. They were accepted (900), then cancelled about 128 s later with 300 Vehicle.request.time.out. With debug logging on, I could see the real reason arriving ~100 ms after each request:

process_code: "901" — [process.901.vehicle.asleep.key]
It failed the same way from the official MyPeugeot app, and even while the car was plugged in and actively charging — so it wasn't a 12V or power issue.

The actual cause, straight from the MyPeugeot app (there's an ⓘ button behind "connection to the vehicle is impossible"):

Your vehicle has not been used for more than 15 consecutive minutes in the last 7 days. To preserve your 12V battery, connectivity is disabled. To use your commands again, take a new 15-minute trip.

It matches my data exactly. My last trips before the outage were 4 min, 13 min 21 s and 9 min 20 s — none reached 15 minutes (the longest missed it by 99 seconds). Looking at my trip export, my last qualifying trip was 33 minutes on 29 July, and connectivity died exactly 7 days later. I'd also made two 14-minute trips in between — 60 seconds short each time, which changed nothing. A 20-minute drive fixed everything instantly.

Why I'm posting this

The rule is mentioned in a comment on #159 and in #73, and #435 is exactly this signature (901 then 300 at +145 s, closed by the stale bot with no reply), but it's nowhere in the README. So people — me included — spend days assuming the integration is at fault. I'd guess it applies to the other Stellantis brands too, not just Peugeot.

Two suggestions:

a troubleshooting note in the README. That alone would probably close a few issues before they're opened.
the integration could likely predict it. The last-trip sensor already exposes an exact duration. If no trip has reached 15 minutes in the last 7 days, it could warn before connectivity drops instead of leaving people to discover it the hard way. I built this on my side with a template sensor and it works well, but it feels like it belongs in the integration.
[...]

PR #529, opened 2026-08-16
Fix: No longer ignore result_code "901" (vehicle asleep) and correct related translations

MoellerDi:
As described in #527 (and confirmed by my own observations), the vehicle appears to enter "deep sleep" if it has not been moved for more than 15 minutes over the past 7 days (the exact duration is secondary; the crucial point is simply that this deep-sleep state occurs).

To make this state more transparent to users, we should display the corresponding wake-up commands sent from the Stellantis backend servers to the vehicle. If the vehicle is indeed in deep sleep (having not moved for some time), this wake-up command times out (a scenario we already track in the command history).

This PR also updates the EN.json and DE.json files accordingly. (EDIT: it covers all *.json files now)

This PR has *not* yet updated the README.md (specifically the error messages section) to explain the situation to users. Please let me know if I should include this in the PR as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,10 @@
 - [Card: Stellantis Vehicles](#card-stellantis-vehicles)
 - [Global preferences](#global-preferences)
 - [Errors](#errors)
+  - [OTP error - NOK:MAXNBTOOLS](#otp-error---nokmaxnbtools)
+  - [OTP error - NOK:NOK_BLOCKED](#otp-error---noknok_blocked)
+  - [Get oauth code error](#get-oauth-code-error)
+  - [Command status 901 and 300 - Wakeup timeout](#command-status-901-and-300---wakeup-timeout)
 - [ABRP - A Better Routeplanner](#abrp---a-better-routeplanner)
 - [Support the project](#support-the-project)
 
@@ -228,6 +232,17 @@ It seems that this error is due to reaching the limit of wrong PIN used. Re-auth
 As described in the "OAuth2 Code > [Remote service](#remote-service)" section, this free service has usage limitations.  
 If you've hit these limits, please wait and try again.  
 If the problem persists or is unrelated to usage limits, please use "OAuth2 Code > [Manual](#manual)" mode.
+
+### Command status 901 and 300 - Wakeup timeout
+**901 (Wakeup Vehicle)** is the wakeup command being forwarded from the Stellantis backend to your vehicle, it doesn't say anything about the vehicle's state by itself. If it's followed by **300 (Timeout)** instead of **0 (Complete)**, the vehicle didn't respond, which means it's in a deep sleep.
+
+Stellantis disables connectivity to protect the 12V battery, the vehicle apps explain it like this:
+
+> Your vehicle has not been used for more than 15 consecutive minutes in the last 7 days. To preserve your 12V battery, connectivity is disabled.
+
+This is tracked per vehicle, not per integration: any trip shorter than 15 minutes doesn't count, no matter how close it gets. If all your trips in the last 7 days were under 15 minutes, connectivity gets disabled and every wakeup times out until a trip passes the threshold again.
+
+To avoid repeated timeouts, take at least one trip longer than 15 minutes every 7 days.
 
 ## ABRP - A Better Routeplanner
 Get a token from [ABRP](https://abetterrouteplanner.com/):
--- a/custom_components/stellantis_vehicles/stellantis.py
+++ b/custom_components/stellantis_vehicles/stellantis.py
@@ -813,8 +813,7 @@ class StellantisVehicles(StellantisOauth):
                         _LOGGER.debug(f"Fetch updates after code: {result_code}")
                         self.do_async(coordinator.async_refresh(), 10)
 
-                    if result_code != "901":  # Not store "Vehicle as sleep" event
-                        self.do_async(coordinator.update_command_history(data["correlation_id"], result_code))
+                    self.do_async(coordinator.update_command_history(data["correlation_id"], result_code))
                 else:
                     _LOGGER.error("No result code")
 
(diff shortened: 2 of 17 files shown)


kZou13:
Thanks for picking this up — this is exactly what I was after in #527.

**README: yes please, I think it's the most useful part.** Showing the `901` only helps if people know what it means. In my case MyPeugeot itself spelled it out: *"Your vehicle has not been used for more than 15 consecutive minutes in the last 7 days. To preserve your 12V battery, connectivity is disabled."* My last three trips were 4, 13 and 9 minutes — the longest missed the threshold by 99 seconds, and connectivity died exactly 7 days after my last 15-minute trip, twice in a row. A short troubleshooting note would save people the two days I spent assuming the integration was broken.

**`300` → `Timeout`: good catch,** the backend literally returns `[Vehicle.request.time.out]`. Worth noting the other 12 language files still say "Canceled" / "Abgebrochen" / etc., so they'd now be less accurate than EN/DE.

**One doubt on `901` → `Wakeup Vehicle`.** All 14 translations currently describe a *state* ("Vehicle as sleep", "Véhicule en veille", "Il veicolo dorme"). Rendered in the command-status sensor it becomes `Start charge: Wakeup Vehicle`, which reads like a step in progress rather than the reason the command failed. `Vehicle asleep` told me immediately what was wrong. Just my two cents — you know the backend semantics better than I do.

I won't submit translations myself since I don't speak most of those languages and I'd rather not guess.

MoellerDi:
> Worth noting the other 12 language files still say "Canceled" / "Abgebrochen" / etc., so they'd now be less accurate than EN/DE.

For the initial pull request, I had only used `en.json` and `de.json` at first, but today I pushed another commit that essentially covers all languages.

> One doubt on 901 → Wakeup Vehicle

Yes, the original definition was intended more as a status, something like "vehicle in sleep mode" ("Vehicle as sleep," "Véhicule en veille," "Il veicolo dorme," etc.). I don't have a strong preference for one variant over the other, but code 901 used to be hidden; and for troubleshooting purposes, it might be more helpful if users can identify which action the subsequent code 300 (timeout) refers to.

Can you confirm that the erroneous sequence looked like this:
`"900": "Accepted" -> "901": "Wakeup Vehicle" -> "300": "Timeout"`?
Or was there also a `"903": "Forwarded"` between 901 and 300?
If so, it makes sense to retain 901 as a status ("vehicle in sleep mode" / vehicle asleep / etc.). If there was no 903 in between, I would view it more as an "executed action" to better understand what the timeout (300) refers to.

What's your opinion?

MoellerDi:
I just encountered a timeout on my eC4 after waking it up. In my case, it looked like this:
`"900": "Accepted" -> "901": "Wakeup Vehicle" -> "300": "Timeout"`
So, at least in my case, there was no `"903": "Forwarded"` between 901 and 300.

image

MoellerDi:
Okay, `README.md` and `info.md` ​​have been updated to explain the situation to users.
In my view, that’s it. Is anything missing, or does anything else need to be changed?

andreadegiovine:

andreadegiovine:
@MoellerDi 
Looking at and using these modifications, I think "901" is a piece of information indicating the vehicle's connectivity status (vehicle asleep); following this, if the vehicle wakes up, we get a 0 (OK), otherwise a 300 (timeout).

What is the purpose of tracking that status?

MoellerDi:
@andreadegiovine 

**Short answer:** `901` tells the user that the vehicle was asleep. That's what puts the following `300` into context: it's a *timeout while waiting for the vehicle to wake up*, not an unexplained cancellation.

**Long answer:**
As described in #527, a vehicle goes into deep sleep after ~7 days without a trip of 15+ minutes. When that happens, every remote command fails, and before this PR the user had no way to tell *why*.

There is no dedicated `wakeup` command. The "wake up" button issues a `state` command, i.e. it asks the backend to pull a fresh status from the vehicle. `901` is the backend telling us "the vehicle is asleep, I'm trying to reach it". Surfacing it makes the whole flow readable.

**Before this PR** a failed `state` command looked like this:

| working | not working |
| --- | --- |
| `900` Accepted | `900` Accepted |
| `903` Forwarded | `300` Canceled |
| `0` Complete | |

The user was left asking: *Why was this canceled? Who canceled it? Is the integration broken?*

**With this PR:**

| working | not working |
| --- | --- |
| `900` Accepted | `900` Accepted |
| `901` Wakeup Vehicle | `901` Wakeup Vehicle |
| `903` Forwarded | `300` Timeout |
| `0` Complete | |

Now the sequence tells the story on its own: the wake-up call was made and it timed out.

The label is the open question, not the value itself. I'm happy to rename `901` to `Vehicle asleep` (as @kZou13 suggested, since it describes the state) or `Waking up vehicle`, whichever you prefer.

andreadegiovine:
The real question is: does the end user need to know about status 901?

We can hide that; I had used the word "canceled" instead of "timeout" simply to let the user know the command hadn't been received by the car. 

We can stick with "timeout" if that conveys the meaning better. 

When looking at the history, constantly seeing "vehicle wakeup" clutters the view; plus, it looks like a command is being sent

—e.g.: Horn -> accepted -> wakeup (sensor= Horn: wakeup) -> completed

What do you think?

MoellerDi:
@andreadegiovine
That works for me. However, @kZou13 complained in #527 that "901" [process.901.vehicle.asleep.key] doesn't show up in the command history. He said that would have saved him a lot of time when troubleshooting...

In any case, I agree with you: if we stick with "timeout" instead of "canceled," that should work well. Should I take care of it, or will you handle it?

andreadegiovine:
I can do it, but im waiting some days for @kZou13 answer to understand why he need this information if the only important notice is command sended or not

PR #532, opened 2026-08-16
Fix: update import statements from Cryptodome to Crypto for consistency (PyCryptodome vs PyCryptodomex)

MoellerDi:
PyCryptodome and PyCryptodomex provide the same core cryptographic code from the same developer. The only difference lies in the namespace: pycryptodome uses the import folder `Crypto`, whereas pycryptodomex uses `Cryptodome` to allow it to coexist alongside older pycrypto packages without file conflicts.

Key differences:
- Namespace: pycryptodome imports via `from Crypto.Cipher import AES`; pycryptodomex imports via `from Cryptodome.Cipher import AES`.
- Coexistence: pycryptodome conflicts with old pycrypto packages; pycryptodomex avoids conflicts and can coexist on the same system.
- Ecosystem usage: pycryptodome is widely used as a standard dependency, whereas pycryptodomex is rarely required—unless a specific legacy system or third-party tool mandates its use.

Home Assistant context:
- Default choice: Home Assistant officially relies on pycryptodome for its core dependencies.
- Conflict warning: Installing or mixing pycryptodome and pycryptodomex in the same Python environment leads to errors caused by slot and file collisions.
- Custom integrations: If a custom component or HACS integration requires pycryptodomex, this can, without proper coordination, interfere with core dependencies or trigger build errors within Home Assistant's restricted container environment.

According to our [manifest.json](https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/blob/d7653083a3c226fc8891ffd14b478495c5beaebf/custom_components/stellantis_vehicles/manifest.json#L17), this integration requires the installation of `pycryptodome==3.*` but imports from `Cryptodome`. In my opinion, this is a discrepancy that this PR is intended to resolve. 

Since Home Assistant already uses [pycryptodome version >=3.6.6](https://github.com/home-assistant/core/blob/22dba261f4b34e7ef72a87051899cf9403f72e98/homeassistant/package_constraints.txt#L83) and enforces this constraint, we should stick with pycryptodome (instead of switching to pycryptodomex) and should uses the correct import folder `Crypto`.

### DIFF
--- a/custom_components/stellantis_vehicles/otp/load.py
+++ b/custom_components/stellantis_vehicles/otp/load.py
@@ -2,7 +2,7 @@ import hashlib
 from locale import atoi
 from time import time
 
-from Cryptodome.Cipher import AES
+from Crypto.Cipher import AES
 
 from .tokenizer import Tokenizer
 
--- a/custom_components/stellantis_vehicles/otp/oaep.py
+++ b/custom_components/stellantis_vehicles/otp/oaep.py
@@ -1,9 +1,9 @@
-import Cryptodome
-from Cryptodome import Random
-from Cryptodome.Cipher.PKCS1_OAEP import PKCS1OAEP_Cipher
-from Cryptodome.Util.number import ceil_div, bytes_to_long, long_to_bytes
-from Cryptodome.Util.py3compat import bord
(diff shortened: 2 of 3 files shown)


andreadegiovine:

Issue #536, opened 2026-08-19
Stellantis servers connections issues


chmtc94:
### Mark

DS

### Model

DS4 E-Tense

### Year

2022

### Engine

Hybrid

### Remote service

Remote control

### HomeAssistant version

HA CORE : 2026.7.4 ; HA OS 18.1

### Integration version

2026.7.2

### What happened?

My vehicle has not moved for 3 weeks. As reported #527, this may cause disconnections from Stellantis server.
I can however observe that :  
- the corresponding entity and icon still indicate connected.
- the log indicate successful calls to the Stellantis server each time_interval with constant returned values. Constant values is normal as the vehicle do not move but calls never fail as well as all token updates...
- each time when i restart HA or when i try to wake-up the vehicle, i get errors in the logs when trying to connect to MQTT server and subscriptions also failed with error 404. Has the URL changed ?
- I updated my local files with PR #529 and #530 for possible better evaluation and provide the corresponding last HA restart log file that show same issues (HA restart : 19-08-2026 12:23:56 and Wake-up try 19-08-2026 13:37:36)

In hollidays without my car, I will be able to move it only next week and will see if something changes...
 
Hope attached log file could help... 

[home-assistant_2026-08-19T11-52-04.518Z_anonymized.log](https://github.com/user-attachments/files/31225181/home-assistant_2026-08-19T11-52-04.518Z_anonymized.log)

EDIT : Error 404 appear only after restart (12:26:25) and concern getting trips and not MQTT subscriptions..

### Log output

```shell

```

chmtc94:
I can understand that:
- communication between the car and the servers is being cut off by the car itself to preserve the auxiliary battery.
- communication between the integration (and/or the MyDS app) and the Stellantis servers remains possible—and thus active at every `update_interval`—but the data on the server is no longer being updated, which explains why the retrieved data remains static.

However, the Stellantis server being queried periodically ought to be able to detect and report that communication with the vehicle has been lost.

This does not however explain the issue with the MQTT server and the integration's failure to subscribe to the two topics. In my view, it would be appropriate to handle these errors as a "disconnected" status with a proper notification from the integration, rather than just a silent log entry that only surfaces when an attempt is made to send a command—such as a wake-up request.

MoellerDi:
Hello @chmtc94, I hope you're doing well and enjoying your vacation :-)
If you have time, could you please test PR #537 and let me know if it works for you?

[...]

PR #537, opened 2026-08-20
Improvement of `get_vehicle_last_trip` and other related methods to better handle cases where no trips are found

MoellerDi:
This PR aims to ~~partially~~ _fully_ resolve issue #536 ~~(the "no trips found" part)~~ by...
- Improving engine status check and avoid fetching last trip data during startup phase after a restart of Home Assistant
- `do_async` gained a wait=False option so callers can fire-and-forget instead of blocking on the future's result
- Improvement of error handling and logging for MQTT subscriptions in `_on_mqtt_subscribe` method.
- Improving the `get_vehicle_last_trip` method to handle cases where no trips are found as well as better pagination and improved timestamp formatting.
- Also further improving general error handling in make_http_request method as detailed below:
  - Fixed a KeyError bug: `result["code"]` → `str(result.get("code"))` for the 404 check, so a 404 response body without a code field no longer crashes with a misleading KeyError instead of surfacing the real error.
  - Fixed a silent-failure bug: previously, any non-2xx status that didn't match one of the specific elif branches (e.g. a plain 403 or 422) fell through and the function returned the response body as if it had succeeded. Now every non-2xx status raises ComunicationError, unless it has already been handled otherwise.
  - Restructured the status branches to properly nest under the "not 2xx" check (they were separate top-level ifs before, which worked by luck since the status codes didn't overlap, but was fragile).
  - Added exception chaining (raise ... from e) so tracebacks show the original cause.
  - Added a dedicated except (ConfigEntryAuthFailed, ComunicationError) branch so those expected/handled errors don't get an extra spurious "Error: ..." warning log on top of the detailed debug logging already emitted above.
   - And added a docstring :-)

This PR was developed and tested on my HA instance and works for me. However, a dedicated pre-release version for further testing would be desirable.

### DIFF
--- a/custom_components/stellantis_vehicles/base.py
+++ b/custom_components/stellantis_vehicles/base.py
@@ -281,11 +281,13 @@ class StellantisVehicleCoordinator(DataUpdateCoordinator):
             if "switch_abrp_sync" in self._sensors and self._sensors.get("switch_abrp_sync") and "text_abrp_token" in self._sensors and len(self._sensors.get("text_abrp_token")) == 36:
                 await self.send_abrp_data()
 
-        if "engine" in self._sensors and "ignition" in self._data and "type" in self._data["ignition"]:
-            current_engine_status = self._sensors.get("engine")
-            new_engine_status = self._data["ignition"]["type"]
-            if current_engine_status != "Stop" and new_engine_status == "Stop":
-                await self.get_vehicle_last_trip()
+        current_engine_status = self._sensors.get("engine")
+        new_engine_status = self._data.get("ignition", {}).get("type")
+        if new_engine_status == "Stop" and current_engine_status not in (None, "Stop"):
+            _LOGGER.debug("Engine status changed from %s to %s, fetching last trip data", current_engine_status, new_engine_status)
+            await self.get_vehicle_last_trip()
+        else:
+            _LOGGER.debug("Engine status unchanged or not relevant for trip data fetch: %s vs %s", current_engine_status, new_engine_status)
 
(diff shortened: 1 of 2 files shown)


MoellerDi:
Further improvements: `do_async` has been extended with a `wait=False` option, allowing callers to use a "fire-and-forget" approach instead of waiting for the `Future` object's result; additionally, `_on_mqtt_subscribe` now uses `wait=False` for delayed reconnection. This prevents the `paho-mqtt` network thread from being blocked—a situation that would otherwise freeze it for the full 300 seconds (resulting in no pings and no processing of other callbacks or data traffic).

MoellerDi:
Another improvement: The integration is checked within `after_async_update_data`:
`if "engine" in self._sensors and "ignition" in self._data and "type" in self._data["ignition"]`. However, it turned out that `self._sensors.get("engine")` returns `None` during the very first run; consequently, `self.get_vehicle_last_trip()` was always executed during the initial data update (e.g., after an HA restart). This has now also been fixed with this PR.

chmtc94:
@MoellerDi 

FYI:  Back home, I simply started my DS4 and then turned it off a few minutes later without moving it. The MyDS app connected successfully to the vehicle, the "waiting for activation" pop-up disappeared, remote controls became available again, and the Home Assistant logs also showed successful vehicle data updates and successful subscriptions to both MQTT topics => the incident—clearly caused by a (too) long period of vehicle inactivity—was resolved with much better information with this PR. 👍

For the "No trip found" message, it reappeared only once—specifically on that occasion—because the vehicle was started and stopped without being moved. The warning in the log for a start/stop cycle without movement might not be relevant. 🤔

andreadegiovine:

PR #540, opened 2026-08-22
Fix: config entry migration skipping steps

MoellerDi:
`async_migrate_entry` migrates old config entries through several steps (versions 1.2 → 1.6). Each step is supposed to save the config entry with that step's target version and minor version. Instead, every step was saving the entry with `StellantisVehiclesConfigFlow.VERSION / .MINOR_VERSION`, which are the integration's current version numbers, not the step's target.

This meant that as soon as the first eligible migration step ran, the entry's stored version jumped straight to the newest version. Every later step checks if `config.version == 1 and config.minor_version < N`, so once the version was bumped early, those checks stopped matching and the remaining steps (folder cleanup, data restructuring, OAuth field migration) were silently skipped — even though the entry was marked as fully migrated.

This PR gives each migration step its own explicit target_version / target_minor_version and saves the entry with those values, so each step only advances the entry by exactly one step and later steps still get a chance to run.

Also added a template block for the next migration step (bump to the current date-based version scheme), to make it easier to add future migrations consistently.

### DIFF
--- a/custom_components/stellantis_vehicles/__init__.py
+++ b/custom_components/stellantis_vehicles/__init__.py
@@ -115,8 +115,10 @@ async def async_remove_entry(hass: HomeAssistant, config: ConfigEntry) -> None:
 
 
 async def async_migrate_entry(hass: HomeAssistant, config: ConfigEntry):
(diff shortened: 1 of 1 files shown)


andreadegiovine:

Issue #544, opened 2026-08-25
Mixed naive/aware updatedAt timestamps silently freeze all sensor updates (TypeError swallowed by except Exception)


krishna3554:
## Summary

`StellantisCoordinator._async_update_data()` compares the previous and new `updatedAt` timestamps to decide whether the API returned fresher data:

```python
current_dt = datetime.fromisoformat(self._data["updatedAt"])
new_dt = datetime.fromisoformat(new_data["updatedAt"])
...
if new_dt <= current_dt:
    return   # skip
```

Two failure modes combine here:

1. If either timestamp carries a UTC offset (`...Z` or `+00:00`) and the other does not, the comparison raises `TypeError: can't compare offset-naive and offset-aware datetimes`. Only `ValueError` is caught by the inner handler.
2. That `TypeError` then escapes to the outer `except Exception: pass`, which discards it - so **`self._data = new_data` never executes** and every entity on the coordinator keeps serving stale values indefinitely.

Static-analysis finding against current `master`; not executed here.

## Location

- File: `custom_components/stellantis_vehicles/base.py`
- Function: `_async_update_data()` (~lines 54-76); outer blanket handler at ~lines 75-76.

## Problem

- Python's `fromisoformat` yields an offset-*aware* datetime when the string ends in `Z`/`±hh:mm` and a *naive* one otherwise. Stellantis/Kamereon payloads are not guaranteed consistent in this respect across endpoints/vehicles (and any past/future format change flips it).
- Mixed awareness makes `<`/`<=` raise `TypeError`, which is outside the `except ValueError` guard.
- The outer `except Exception: pass` was presumably meant to shield transient API errors, but it converts this persistent condition into a silent, permanent stall: each poll re-fetches fresh data, fails the same way, and throws the fresh payload away.

## Trigger / Reproduction

Based on static analysis; no runtime run performed:

1. Coordinator has cached data whose `updatedAt` is e.g. `"2026-08-01T12:00:00+00:00"` (aware).
2. API returns newer data with `"2026-08-02T12:00:00"` (naive) or vice versa.
3. Comparison raises `TypeError`; outer handler swallows it; sensors freeze at the old values while logs show nothing above DEBUG level.

## Expected Behavior

[...]

PR #545, opened 2026-08-25
Fix error handling and timezone comparison in _async_update_data

MoellerDi:
This PR fixes the situations described in #544.

- Normalize `updatedAt` timestamps to UTC-aware `datetime` objects before comparing them, fixing a `TypeError: can't compare offset-naive and offset-aware datetimes` that could be raised when one timestamp carries a UTC offset (`Z` / `+00:00`) and the other doesn't.
- Catch `TypeError` alongside `ValueError` when parsing `updatedAt`, so a non-string value degrades gracefully (skip the staleness comparison, proceed with the update) instead of being swallowed further up.
- Replace the bare `except Exception: pass` with `raise UpdateFailed(...)`, so unexpected API errors (timeouts, connection issues, 5xx, unexpected responses) are properly surfaced to the coordinator instead of being silently discarded. This lets `last_update_success` and entity availability reflect real failures, and lets Home Assistant log and retry as designed.
- Keep the raised `UpdateFailed` message generic and log the actual exception only at debug level, avoiding exposure of exception details that could (in some error paths) include request URLs with sensitive query parameters.

### DIFF
--- a/custom_components/stellantis_vehicles/base.py
+++ b/custom_components/stellantis_vehicles/base.py
@@ -4,7 +4,7 @@ from datetime import datetime, timedelta, UTC
 import json
 from copy import deepcopy
 
-from homeassistant.helpers.update_coordinator import ( CoordinatorEntity, DataUpdateCoordinator )
+from homeassistant.helpers.update_coordinator import ( CoordinatorEntity, DataUpdateCoordinator, UpdateFailed )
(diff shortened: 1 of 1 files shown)


andreadegiovine:
[2026.8.1-beta.2](https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/releases/tag/2026.8.1-beta.2)

Issue #550, opened 2026-08-26
Two vehicle management issue


chmtc94:
### Mark

DS

### Model

N°4

### Year

2025

### Engine

Hybrid

### Remote service

Connect Plus

### HomeAssistant version

 Core 2026.7.4 Supervisor 2026.07.5 Operating System 18.2 Interface utilisateur 20260624.6

### Integration version

2026.8.1-Beta.2

### What happened?

Following up on discussion #543, I finally managed (after 4 weeks!) to activate connected services for a second vehicle in the MyDS application (same user email, same password, same security code).

I restarted Home Assistant to check if the integration had recognized the second vehicle. (See attached full log)

In the Home Assistant integration tab, the second vehicle has been successfully automatically added, and when the vehicle is selected, the various dashboards for this vehicle appear with property values ​​that correctly match those of the vehicle (electric range, fuel, etc.). I was thus able to create a dedicated view grouping all the data associated with this second vehicle.
Everything seems to be working perfectly. I will be able to test charging the vehicle tonight.

However, the log shows:

1) At the time of the restart (line 42 of the log), a 404 error occurred immediately after the request to retrieve the vehicle list.

I have attached the retrieved data in JSON format (replace single quotes by double quotes, True by true, False by false and None by none)  which is the result of the request, but this may not be the cause...

[vehicles-anonymized.json](https://github.com/user-attachments/files/31486128/vehicles-anonymized.json)

EDIT : Now that the integration is stable enough, it could be interesting to check some listed requests in the above file, perhaps as services 🤔

2) Vehicle 2 has its own data update interval, which may therefore differ from that of vehicle 1. When vehicle 2 was created, at the time of the HA restart, the interval was set to 60 seconds. However, the attempt to change this interval from 60 seconds to 90 seconds around 6:30 PM resulted in a slight delay in the data update request. The new interval was not applied and remained at 60 seconds (with the "update interval" entity correctly remaining at 90 seconds).

<img width="568" height="627" alt="Image" src="https://github.com/user-attachments/assets/db5e8d9d-b2fc-4e00-8409-4a49331fe999" />

We might also question the relevance of having two periods and a possible way to desynchronize update requests between vehicles to avoid, as much as possible, bottlenecks at the server level (for example, by ensuring a staggered response time between requests from the two (or more) vehicles (a precaution that may be completely unnecessary, however?).

[...]

PR #552, opened 2026-08-27
Improve vehicle picture download failure log message

MoellerDi:
Just a small change to improve a single error message. See #550 for details (the part regarding the 404 error message).


chmtc94:
I can confirm that no picture is present in the provided stellantis card. (One is present in the MyDS App)

For the first vehicle there is a png file of the vehicle stored in config/www/stellantis-vehicle/<ID>/<VIN_1>.png

Is it the right place where the <VIN_1>.png file for the second file should have been downloaded? 

May it be suffisant to place here such a file directly but if yes with which resolution?

MoellerDi:
https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/blob/8974d0ae8498ba32c8418fd50cfc36e86a93c5bb/custom_components/stellantis_vehicles/stellantis.py#L507

Issue #216, opened 2025-06-29
Feature request: New sensors, driving mode and others


Jordan87:
### Mark

Peugeot

### Model

308 PlugInHybrid

### Year

2023

### Engine

Hybrid

### Remote service

E-remote control

### HomeAssistant version

2025.6.3

### Integration version

2025.6.7-beta.4

### What happened?

Today I was checking what kind of data is also provided from stellantis server and found few interesting information.

1. drivingBehavior - mode, which in my case with PlugInHybrid was providing values:
- "ZEV" - only EV mode
- "Hybrid"
- "Sport"
- "Normal" - reported when parked
'drivingBehavior': {'createdAt': '2025-06-28T11:47:54Z', 'mode': 'ZEV'}

2. Doors and windows state
In log there is a lot of detailed info. Currently we have only lock state, for doors maybe opening state won't be very useful, but to have windows state it could be very useful.
'doorsState': {'createdAt': '2025-06-27T15:08:27Z', 'lockedStates': ['Locked', 'RearDoorsLocked'], 
'opening': [{'identifier': 'Trunk', 'state': 'Closed'}, {'identifier': 'Driver', 'state': 'Closed'}, {'identifier': 'RearRight', 'state': 'Closed'}, {'identifier': 'RearLeft', 'state': 'Closed'}, {'identifier': 'Passenger', 'state': 'Closed'}, {'identifier': 'RearWindow', 'state': 'Closed'}, {'identifier': 'RoofWindow', 'state': 'Open'}]}

3. Belt status, potentially interested but at least in EU annoying sound is used if driver or front passenger is not fastened so I'm not sure if it is even possible to dive without it :P
'safety': {'createdAt': '2024-07-18T13:48:49Z', 'beltStatus': [{'belt': 'Normal', 'id': 'Driver'}, {'belt': 'Omission', 'id': 'Passenger'}]}

4. Lightning system - turn signal and fog lights, probably not useful but interesting that is is also provided :)
[...]

PR #553, opened 2026-08-27
Add driving behavior sensor

MoellerDi:
Minor change to support #216


andreadegiovine:
Please add italian translation:

"Modalità di guida"

PR #558, opened 2026-08-29
Improvement: Staggered execution of periodic API polls across multiple vehicles

MoellerDi:

## As of today

`async_setup_entry` creates every vehicle's coordinator and runs its first
refresh back-to-back, all with the same `UPDATE_INTERVAL`. The coordinators
therefore stay phase-locked and, on a multi-vehicle account, poll the Stellantis
API for all vehicles at (nearly) the same instant on every cycle. This bursts
the API and makes debug logs harder to read (per-vehicle events are interleaved).

## Change

- `__init__.py::async_setup_entry`: `enumerate(vehicles)`; after the first
  refresh, vehicle *i* (`i > 0`, and only when there is more than one vehicle)
  gets a one-time phase shift of `i * UPDATE_INTERVAL / len(vehicles)` seconds
  via `coordinator.stagger_first_poll(...)`.
- `base.py`:
  - `stagger_first_poll(offset_seconds)` pushes the next poll back once by
    lengthening `update_interval` for a single cycle.
  - `_async_update_data` consumes the offset on the next run and restores the
    normal cadence (a `number_refresh_interval` override is re-applied right
    after, unchanged).

Only the phase changes; the polling rate is unchanged, and per-vehicle `number_refresh_interval` still works.

## Behaviour

| | Before | After (2 vehicles) |
|---|---|---|
| Poll timing | both vehicles polled together every cycle | ~30 s apart |
| After restart | phase-locked again | re-staggered automatically |

### DIFF
--- a/custom_components/stellantis_vehicles/__init__.py
+++ b/custom_components/stellantis_vehicles/__init__.py
@@ -18,7 +18,8 @@ from .const import (
     INTEGRATION_VERSION,
     PLATFORMS,
     OTP_FILENAME,
-    FIELD_NOTIFICATIONS
+    FIELD_NOTIFICATIONS,
+    UPDATE_INTERVAL
 )
 
 _LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


MoellerDi:

andreadegiovine:
@MoellerDi if it work and solve the multi vehicles problems (i have only one vehicle) without problem for single vehicles, we can merge it 😀

Around 30/40% of integration improvments came from you

MoellerDi:
Go for it :-) I'll check the others for conflicts once this one is merged.

PR #561, opened 2026-08-29
Improvement: Detect an unpaired vehicle without a restart

MoellerDi:

## Problem

When a vehicle is removed from the Stellantis account, its `/status` endpoint returns `404 / 40400` and the integration keeps polling and warning about it until Home Assistant is restarted. Only the restart re-runs `get_user_vehicles` and notices the vehicle is gone. There is no runtime reconciliation.

## Root cause

`get_user_vehicles` caches the account vehicle list in `self._vehicles` and never refetches it for the lifetime of the config entry (`if not self._vehicles:`).

## Change

- `get_user_vehicles(force=False)`: with `force=True` the cache is dropped and the account vehicle list is fetched again.
- `StellantisVehicleCoordinator`:
  - counts consecutive empty vehicle-status responses;
  - after `EMPTY_STATUS_LIMIT` (3) it calls `get_user_vehicles(force=True)` once and, if this vehicle's VIN is no longer on the account, raises a repair issue (`vehicle_removed`) and stops re-checking;
  - if the list still contains the VIN (or could not be fetched) it resets the counter and keeps polling;
  - a later non-empty response clears the issue and logs the recovery once.
- `EMPTY_STATUS_LIMIT` added to `const.py`; `issues.vehicle_removed` added to `translations/en.json` (other locales left to their maintainers).

This is intentionally conservative: it only surfaces the removal (repair issue + entities go stale). It does **not** unsubscribe MQTT, tear down the coordinator or delete the device/entities at runtime – an HA restart still performs that cleanup.

## Behaviour

| | Before | After |
|---|---|---|
| Removal noticed | only on next HA restart | within a few poll cycles |
| User signal | none | Repairs issue with removal instructions |
| Extra API calls | none | one vehicle-list fetch after 3 empty status responses, then none until it recovers |

## Scope / non-goals

- No runtime teardown of MQTT / coordinator / registry entries.
- Reacts to *empty* status responses; it does not itself change how an empty response is handled elsewhere.

### Repair
image

### Repair Details
image

### DIFF
--- a/custom_components/stellantis_vehicles/__init__.py
+++ b/custom_components/stellantis_vehicles/__init__.py
@@ -5,7 +5,7 @@ import os
 from homeassistant.core import HomeAssistant
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.exceptions import ConfigEntryAuthFailed
-from homeassistant.helpers import issue_registry
+from homeassistant.helpers import issue_registry, device_registry as dr
 from homeassistant.components.frontend import add_extra_js_url
 from homeassistant.components.http import StaticPathConfig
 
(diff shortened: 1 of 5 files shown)


chmtc94:
@MoellerDi 
I found, You fixed 😉👍

EDIT 1 :  However, deleting the device will almost certainly also delete all its entities and, consequently, all its short- and long-term statistics.
That is what I had in mind when considering saving data for each charge (`last_charge_sensor`) and each trip (`last_trip_sensor`) to a file or database table (e.g., .csv, .json, .xls, .sqlite). This is similar to how the MyDS app handles trips (using a .myd JSON file containing all trip data)—which is actually one of the few things it does reasonably well, though even then, some trips still end up being lost forever.
In my opinion, the only two things worth archiving for the medium to long term are trips and charging sessions. If we can store this data continuously as it occurs, there would be no downside to deleting the devices and all associated entities.

EDIT 2 : However, regarding trips, this backup might not be necessary, provided they can all be retrieved from the corresponding menu in the MyDS app (Driving/Trips/Save your trips) before removing the vehicle from the account. 
A dedicated integration service could also handle this task using the get_trips request

Therefore, only charging sessions would really be worth archiving...

MoellerDi:

MoellerDi:
@chmtc94 
>However, deleting the device will almost certainly also delete all its entities and, consequently, all its short- and long-term statistics.

True, while the user can still "ignore" this repair notification, the final decision rests with them. Perhaps we should make it clearer in the text that all entities, and thus all short- and long-term statistics associated with them, will also be removed...

chmtc94:
> True, while the user can still "ignore" this repair notification, the final decision rests with them. Perhaps we should make it clearer in the text that all entities, and thus all short- and long-term statistics associated with them, will also be removed...

And what about a desactivation instead of a deletion ? 
I don't really have a firm grasp of the difference in consequences...

MoellerDi:
> > True, while the user can still "ignore" this repair notification, the final decision rests with them. Perhaps we should make it clearer in the text that all entities, and thus all short- and long-term statistics associated with them, will also be removed...
> 
> And what about a desactivation instead of a deletion ? I don't really have a firm grasp of the difference in consequences...

Sure, I mean the user can just ignore this repair and just disable it in the settings. No problem 😏

MoellerDi:
I will resolve the conflicts this evening or by tomorrow at the latest. Feel free to go ahead and merge the others that don't have any conflicts.

MoellerDi:
All conflicts resolved

MoellerDi:
All conflicts resolved (again)

MoellerDi:
@andreadegiovine All conflicts resolved (again)

PR #585, opened 2026-09-05
Use the API's brand field for the device manufacturer

MoellerDi:
## Summary

The Stellantis connected-car API never returns a human-readable vehicle model name — only VIN, `motorization` (engine type) and `brand`. `StellantisBaseEntity.device_info` (in `base.py`) worked around this by showing the translated engine type + VIN as `model`, and the user's selected mobile app (`FIELD_MOBILE_APP`) as `manufacturer`, discarding the API's own `brand` field entirely.

There is no public decode table for Stellantis' VDS (VIN positions 4-8), so a real model name can't be derived automatically. Instead, this PR adds a small manually-curated VIN lookup that improves `manufacturer` immediately (using data the API already sends) and lets `model` be filled in as users confirm their own vehicle's VIN prefix against their registration document.

### 1. New `vin.py`

- `WMI_BRANDS`: VIN positions 1-3 (World Manufacturer Identifier) → brand, seeded with known Stellantis-brand codes. Used only as a fallback when the API's own `brand` field is unavailable.
- `VIN_MODELS`: VIN positions 1-8 (WMI + VDS) → model name. Empty until a user confirms an entry; a few were seeded upfront, cross-checked against the free NHTSA vPIC API where possible (NHTSA has no data for EU-market Peugeot/Citroën/Opel/DS VINs — expected, not a bug).
- `get_brand_from_vin(vin)` / `get_model_from_vin(vin)` helpers.

### 2. `stellantis.py`: capture the API's `brand` field

`vehicle_data` (built from `_embedded.vehicles` in `get_user_vehicles`) now keeps `vehicle.get("brand")` instead of discarding it.

### 3. `base.py`: `device_info` fallback chain

- `manufacturer`: `vehicle["brand"]` (from the API) → `get_brand_from_vin(vin)` → `FIELD_MOBILE_APP` (previous behaviour), in that order.
- `model`: `get_model_from_vin(vin)` if a confirmed entry exists, else the previous `"{engine label} - {vin}"` display.

## Behaviour

- `identifiers` in `device_info` are unchanged, so no new device is created — existing devices get `manufacturer`/`model` updated in place the next time entities are (re-)added (integration reload or HA restart).
- Vehicles with no matching `VIN_MODELS` entry, or whose API response has no `brand`, behave exactly as before.

## Scope / non-goals

- No automatic full VIN decoding — Stellantis doesn't publish the VDS table, so `VIN_MODELS` only grows from confirmed entries.

## Manual verification and testing against a running instance: 
- [x] reload the integration and confirm the device card shows the expected `manufacturer` (from the API's `brand` field) and, for vehicles matching a `VIN_MODELS` entry, the expected `model` — with no duplicate device created.

### DIFF
--- a/custom_components/stellantis_vehicles/base.py
+++ b/custom_components/stellantis_vehicles/base.py
@@ -483,7 +483,7 @@ class StellantisBaseEntity(CoordinatorEntity):
             },
             "name": self._vehicle["vin"],
             "model": self._coordinator.get_translation(f"component.stellantis_vehicles.entity.sensor.type.state.{self._vehicle["type"].lower()}", self._vehicle["type"]) + " - " + self._vehicle["vin"],
-            "manufacturer": self._config[FIELD_MOBILE_APP]
+            "manufacturer": self._vehicle.get("brand") or self._config[FIELD_MOBILE_APP]
         }
 
     def value_was_updated(self):
--- a/custom_components/stellantis_vehicles/stellantis.py
+++ b/custom_components/stellantis_vehicles/stellantis.py
@@ -677,7 +677,8 @@ class StellantisVehicles(StellantisOauth):
                         vehicle_data = {
                             "vehicle_id": vehicle["id"],
                             "vin": vehicle["vin"],
-                            "type": vehicle["motorization"]
+                            "type": vehicle["motorization"],
+                            "brand": vehicle.get("brand")
                         }
                         try:
                             picture = await self.resize_and_save_picture(vehicle["pictures"][0], vehicle["vin"])


andreadegiovine:
Great work; I’ve always wanted to avoid calculating the data myself—it’s better to just use the data received from the APIs.

If we add this, many users will open issues asking to add vehicles (look at PSA).

MoellerDi:
> If we add this, many users will open issues asking to add vehicles (look at PSA).

I agree with you. I will revise this PR so that it only uses what we receive from the API (mainly `vehicle["brand"]`).

MoellerDi:
Thanks for the feedback. You're right that the VIN_MODELS table would just turn into a stream of "add my VIN" requests, so I've dropped the VIN lookup entirely — no more vin.py, no WMI_BRANDS, no VIN_MODELS.

What's left is only the part that uses data the API already sends: get_user_vehicles() receives a brand field per vehicle that was being discarded. The PR now keeps it in vehicle_data and uses it for device_info["manufacturer"], falling back to the mobile app name when the API doesn't provide one.

PR #591, opened 2026-09-06
Add config entry diagnostics

MoellerDi:
## Summary

The integration keeps a lot of non-persistent runtime state that never surfaces anywhere in the UI: OAuth and MQTT token expiry moments, whether a token refresh is even scheduled, the MQTT connection state, per-vehicle coordinator counters (`_empty_status_count`, `_vehicle_removed`, `_disabled_commands`), the command history with its result codes, and the raw API payload. Today, getting any of this out of a user's instance means several rounds of "enable debug logging, reload, wait, anonymize, upload".

Without a `diagnostics.py` the "download diagnostics" button does not exist for this integration. This PR adds one, so a single JSON download carries everything needed to diagnose the common failure reports (entities unavailable, remote commands do nothing, sensor value wrong, config broken after update).

### New `diagnostics.py`

One function, `async_get_config_entry_diagnostics`. No changes to any existing file; no manifest change (the `diagnostics` platform is auto-detected).

Top level:

- `entry`: `version`, `minor_version`, and `data_keys` (key names only, since pre-migration entries can still hold tokens in `entry.data`).
- `config`: curated summary (`mobile_app`, `country_code`, `remote_commands`, `anonymize_logs`, `notifications`) plus the `_config` key list. `_config` is not dumped wholesale: it nests a VIN-keyed `vehicles` sub-dict, and `async_redact_data` masks values by key name, not dict keys, so the VINs would leak.
- `oauth`: `token_expires_at`, `refresh_scheduled`.
- `mqtt`: `enabled`, `connected`, `token_expires_at`, `refresh_token_expires_at`, `refresh_scheduled`, `request_pending_reconnect`.
- `vehicles[]`: one entry per coordinator (see below).

Per vehicle:

- `vin`: first 8 characters (WMI plus descriptor section, so manufacturer and model stay readable) and last 3, middle masked with `*`.
- `type`, `brand`, `update_interval_seconds`, `last_update_success`.
- `last_exception`: `repr()` of the coordinator's last failed refresh, truncated to 500 characters.
- `empty_status_count`, `vehicle_removed`, `disabled_commands`, `commands_history` (via the existing `command_history` property, already formatted and PII-free), `pending_programs` (read with `getattr`, since `develop` does not have that attribute yet).
- `sensors` and `raw_status`: `_sensors` / `_data` run through `async_redact_data(...)` and then `_scrub_substrings(...)` (see Redaction).

### Redaction

`TO_REDACT` covers tokens, `customer_id`, `client_id` / `client_secret` / `basic_token`, `code*`, `pin`, `email` / `password`, `text_abrp_token`, `vin`, `vehicle_id`, `_links`, and GPS (`lastPosition`, `coordinates`, `latitude`, `longitude`). It overlaps with `SensitiveDataFilter.masked_entry_keys` in `utils.py`; the extra keys here cover the raw API payload, which never reaches the log filter.

`async_redact_data` only masks whole values by key. The raw `_status` payload also carries the VIN and the opaque PSA vehicle id embedded inside `_links` URLs and other free-text fields, so `_scrub_substrings` runs afterwards over `sensors` and `raw_status` and replaces those two values (guarded to at least 8 characters) wherever they appear inside a string.

## Behaviour

- Read-only. Nothing is called on the API, no state is changed, no requests are made.
- The button appears in the integration and device pages once the file is present.

## Scope / non-goals

- `async_get_config_entry_diagnostics` only; no `async_get_device_diagnostics` (a single download already carries every vehicle).
- `last_exception` is included verbatim (truncated). A token embedded in an API URL inside an error message is only masked when the user has `anonymize_logs` enabled.

## Manual verification and testing against my running instance: 
- [x] reload the integration, use "download diagnostics", and confirm the JSON contains no full token, no full VIN (only `WMI+VDS***XYZ`), and no coordinates; and that `oauth` / `mqtt` expiry fields, `commands_history`, and the redacted `raw_status` are populated as expected.

### DIFF
--- a/custom_components/stellantis_vehicles/diagnostics.py
+++ b/custom_components/stellantis_vehicles/diagnostics.py
@@ -1,0 +1,165 @@
+"""Diagnostics support for Stellantis Vehicles."""
+
+from __future__ import annotations
+
+from typing import Any
+
+from homeassistant.components.diagnostics import async_redact_data
+from homeassistant.config_entries import ConfigEntry
+from homeassistant.core import HomeAssistant
+
+from .base import StellantisVehicleCoordinator
+from .const import (
+    DOMAIN,
+    FIELD_ANONYMIZE_LOGS,
+    FIELD_COUNTRY_CODE,
+    FIELD_MOBILE_APP,
+    FIELD_NOTIFICATIONS,
+    FIELD_REMOTE_COMMANDS,
+)
+
+# Keys whose value is a secret or personal data. async_redact_data walks nested
+# dicts/lists and replaces every matching value with "**REDACTED**", so only this
+# list needs maintaining. It overlaps with SensitiveDataFilter.masked_entry_keys
+# in utils.py; the extra entries here cover the raw API payload (GPS position,
+# VIN) that never reaches the log filter.
+TO_REDACT = {
+    "access_token",
+    "refresh_token",
+    "id_token",
+    "oauth_code",
+    "code",
+    "code_verifier",
+    "code_challenge",
+    "customer_id",
+    "pin",
+    "email",
+    "password",
+    "vin",
+    "vehicle_id",
+    "text_abrp_token",
+    "lastPosition",
+    "coordinates",
+    "latitude",
+    "longitude",
+    # HAL navigation block: its "href" URLs embed the opaque PSA vehicle id as a
+    # path segment, which async_redact_data cannot reach inside the string.
+    "_links",
+}
+
+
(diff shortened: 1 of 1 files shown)


andreadegiovine:
Same here, client_id secret and basic_token has exposed on this repo, why need to mask it?

MoellerDi:
> Same here, client_id secret and basic_token has exposed on this repo, why need to mask it?

No need; just me and my confusion ;-) Removed and ready for review.

Issue #414, opened 2026-01-22
Intensive processor usage issue


chmtc94:
### Mark

DS

### Model

DS4 E-Tense

### Year

2022

### Engine

Hybrid

### Remote service

E-remote control

### HomeAssistant version

Core 2026.1.0 - Supervisor 2026.01.1 - Operating System 16.3 

### Integration version

2025.12.9

### What happened?

Since some weeks, I encounter a regular intensive processor usage issue.

This issue exists for a long time (also in 2025) but I only tried to find its origin after its occurrence on 2026-01-14. 
As an HA restart (for example each HA Core or HA OS update) solved temporaly the issue, I was not in a hurry and I did not suspect that the issue might come from the Stellantis integration.

1) Example : 10/01/2026 TO 15/01/2026 WITH AN HA CORE UPDATE TO VERSION 2026.1.0 ON 11/01/2026 22:32 TO 22:57

<img width="1210" height="834" alt="Image" src="https://github.com/user-attachments/assets/0b3dbf60-19d2-4f6f-8bed-e6a52b195107" />

Issue detail 11/01/2026 14:00 to 15:00 

<img width="1210" height="834" alt="Image" src="https://github.com/user-attachments/assets/a992ee60-6c51-4a0c-8bd2-56119f3bff74" />

2) FOCUS on 14/01/2026 00H00 - 16/01/2026 09H00

<img width="941" height="503" alt="Image" src="https://github.com/user-attachments/assets/d94037fd-9bda-47e2-90d4-bad39ce0ee0e" />

Detail 14/01/2026 20H00 - 15/01/2026 04H00

<img width="1210" height="834" alt="Image" src="https://github.com/user-attachments/assets/553538b6-a26c-4ff4-ad60-da295d0ec2f6" />

More detail : 14/01/2026 21H00-22H30

<img width="1210" height="834" alt="Image" src="https://github.com/user-attachments/assets/6356cb34-ec01-45e4-a5f4-319e0319a046" />

Following corresponding log file with focus on last line of _async_update_data calls where we can see the call duration : 

[2026-01-14 21H00-22H30.log](https://github.com/user-attachments/files/24807734/2026-01-14.21H00-22H30.log)

[...]

PR #593, opened 2026-09-06
Keep the sensitive-data log filter bounded and per config entry

MoellerDi:
_EDIT: just added another commit, so text below changed to include it as well._

## Summary

`SensitiveDataFilter` is one shared instance attached to the `base` and `stellantis` module loggers. Two problems made it both leak CPU and miss values it should have masked:

- Over the lifetime of a running Home Assistant process its set of masked values only ever grew, which is the steadily rising `_async_update_data` CPU time reported in issue #414 (the time is spent inside the log filter, not in the HTTP call). `refresh_token_request` and `refresh_mqtt_token_request` registered every rotated oauth or mqtt token through `add_custom_value`; each rotation produces genuinely new strings, so the value list kept growing, the compiled mask pattern kept getting longer, and `re.sub` over every debug record kept getting slower.
- With more than one account configured, `add_entry_values` merged each entry's config into a single flat dict. The keys collide (`oauth`, `mqtt`, `customer_id`), so the last entry to call `save_config` overwrote the other accounts' tokens in the mask, and those tokens were then logged unmasked.
- `_mask_value` only handled `str`, `dict`, `list` and `tuple`. MQTT message payloads are logged as raw `bytes` (`_on_mqtt_message`), so a VIN or token inside the JSON payload passed through completely unmasked, even when it was in the compiled pattern. This gap predates issue #414 and is present on `develop`.

This branch carries three commits:

1. `Attach the sensitive-data log filter once per module logger`: the filter used to be rebuilt and re-added to the module loggers on every `StellantisBase.__init__` and every coordinator init, with nothing ever removing them, so filters stacked up on the loggers. It is now a module-level singleton attached exactly once at import.
2. `Scope the sensitive-data log filter per config entry and bound its value set`: per-entry masked values, a bounded extras FIFO, thread-safe rebinds, and the `save_config` / `async_shutdown` wiring (below).
3. `Mask bytes log payloads and register account VINs per entry`: the `bytes` handling, plus a dedicated per-entry store for the account's VINs so they are not evicted by the bounded FIFO (below).

## Change

`utils.py`, `SensitiveDataFilter`:

- Masked values are kept per `entry_id` in `_entry_values`. `set_entry_values(entry_id, entry_data)` replaces that entry's snapshot wholesale, so a rotated token supersedes the previous one and several accounts no longer overwrite each other. It also collects the VINs that are keys of the per-vehicle config node.
- `_entry_extra` holds, per `entry_id`, the account's VINs and vehicle ids from the live vehicle list (`set_entry_extra_values`). It is separate from `_entry_values` so a later config snapshot cannot drop it, and out of the FIFO so it is never evicted while the entry is loaded.
- `remove_entry_values(entry_id)` forgets an entry (both stores) so the compiled pattern shrinks back and that entry's now invalid tokens stop being masked.
- `add_custom_value` is a bounded FIFO (`CUSTOM_VALUES_LIMIT = 128`) for values seen outside the stored config (the OAuth code and id_token during the auth flow). Anything that must stay masked for an entry's lifetime is in `_entry_values` or `_entry_extra`.
- `_mask_value` now also handles `bytes` and `bytearray`: it masks the decoded UTF-8 text and re-encodes, returning the original object unchanged when nothing matched.
- Every mutating method rebinds its container instead of mutating in place, so the filter can be read from the paho-mqtt network thread while the event loop updates it, without a "dictionary changed size during iteration".
- The anonymize check is `any(self._entry_anonymize.values())`. A per-module logger cannot tell which entry a record came from, so masking is on whenever any loaded entry enabled it.

`stellantis.py`:

- `save_config` is the single point through which oauth or mqtt token rotations and per-vehicle settings reach `self._config`, so it refreshes the filter snapshot for the entry (a no-op until `set_entry` has run, that is during the config flow).
- `update_vehicle_stored_config` and `prune_stored_vehicle_configs` now write `_config["vehicles"]` through `save_config({"vehicles": ...})` instead of a direct assignment, so a per-vehicle config change made after setup also refreshes the snapshot.
- `get_user_vehicles` registers the account's VINs and vehicle ids with `set_entry_extra_values` (or, during the config flow when there is no entry yet, the bounded FIFO).
- `async_shutdown` calls `remove_entry_values` for the entry.
- `refresh_token_request` and `refresh_mqtt_token_request` persist the new config before logging the raw response, so the freshly issued tokens are masked through the refreshed snapshot instead of a hand-written `add_custom_value` per rotation. `refresh_mqtt_token_request` keeps one targeted `add_custom_value` in the "no access_token in response" branch, where there is no config to persist.

## Behaviour

| | Before | After |
|---|---|---|
| Long-running process, periodic token refresh | masked value set grows without bound, log filter CPU per record keeps rising | value set is bounded by the loaded entries, constant over time |
| Two or more accounts | last `save_config` wins, other accounts' tokens logged unmasked | each entry's values kept separately, all masked |
| Entry unload or reload | filter kept the removed entry's tokens forever | entry's values dropped, pattern shrinks |
| `anonymize_logs` differing between entries | effectively "last writer wins" | masking on if any loaded entry enabled it |
| VIN or token inside an MQTT payload (`bytes`) | never masked | masked (payload stays `bytes`) |
| VIN of a vehicle with no stored per-vehicle config | ~~only in the FIFO, evicted after roughly a day of token rotations, then reappears unmasked~~ works, but degrades over time | in `_entry_extra`, masked for the entry's whole lifetime |
| Single account, anonymize on, normal run | works, but degrades over time | works, stable |

## Scope and non-goals

- No change to which requests are made, to entity states, or to config entry data.
- The filter still only covers the `base` and `stellantis` module loggers, not `config_flow`, `otp`, or `utils`. Widening its reach is out of scope here.
- No new config option. The existing `anonymize_logs` toggle is untouched; this change makes leaving it on cheap enough that it no longer needs to be worked around.

## Manual verification

- [x] Enable `custom_components.stellantis_vehicles: debug`, keep `anonymize_logs` on, reload the integration. Entities populate, and tokens and VIN appear as `AAAAA###` in the log.
- [x] Wait for an MQTT `state` message. The VIN inside the `b'{"vin":"..."}'` payload is masked.
- [x] Let the integration run through at least one oauth and one mqtt token refresh. The freshly issued tokens are masked in the `_log_http_exchange` line, and the `_async_update_data` duration does not climb over successive updates.
- [ ] With two accounts configured, both accounts' tokens are masked in the log.
- [x] Reload one entry, then unload it. No stray errors, and after unload that entry's former token no longer appears (masked or not) in new log lines.


MoellerDi:
@chmtc94 I saw that you mentioned performance issues occurring when the "Anonymize logs" configuration option is enabled. This PR addresses exactly that, could you please test it if you have the time?

chmtc94:
> @chmtc94 I noticed you reported performance issues when the "Anonymize logs" configuration option is enabled. This PR addresses exactly that issue; could you test it if you have the time?

I’ll test this as soon as possible and get back to you 😉 However, issue #414 only appears suddenly (and significantly) a few days after enabling the "Anonymize logs" option. I fully understand the cause (the ever-growing amount of stored data).

Also, when an OTP code is required, the PIN code can appear in the log and needs to be masked as well—possibly along with the OTP credentials used to obtain it. I mask them manually whenever I spot a PIN code in the log that has already been masked by the script.

Personally, I’ve always kept logging enabled for this integration; I extract the data roughly every four days and keep the extracted log files in a dedicated folder, just in case later analysis proves useful. If you need data from a specific period, feel free to ask 😉

One quick question: does your PR use different masked values ​​for each occurrence of the same key (KEYNAME)—similar to how my personal script #562 uses KEYNAME_i for each occurrence *i* of a key? That seems useful for log analysis, as it allows one to spot whether values ​​are identical or not 😉

MoellerDi:
@chmtc94 
> I’ll test this as soon as possible and get back to you 😉 However, issue https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/issues/414 only appears suddenly (and significantly) a few days after enabling the "Anonymize logs" option. I fully understand the cause (the ever-growing amount of stored data).

Thank you very much for your kind support and for conducting the tests. It is precisely this "steadily growing volume of stored data" that this PR aims to reduce.

> when an OTP code is required, the PIN code can appear in the log

Yes, some time ago I created a PR (#147) that originally [corrected or removed OTP logging in such cases](https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/pull/147/changes/cb8efab751861aec53d278365694e8217fc20ae3#diff-227456f8be2669c737aae1aae8f8fdbc3901bf55d6d5d9431e8efa545d553573). However, I was asked to revert the change at the time, as the goal was to keep the files as close as possible to the original psa_controller version. I am not sure if that still applies.

> One quick question: does your PR use different masked values ​​for each occurrence of the same key (KEYNAME)—similar to how my personal script https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/discussions/562 uses KEYNAME_i for each occurrence i of a key?

Masking is performed (currently) as before: `f"{value_str[:5]}###"`; i.e., a value like "abcdefghij..." becomes "abcde###" in the log.

Btw, this PR is marked as a draft because I anticipate conflicts that I will need to resolve once my other PRs have been merged. So, there is plenty of time for testing... :-)

MoellerDi:
Rebased onto latest develop

MoellerDi:
@chmtc94 

I have also just added masking for the OTP PIN code to this PR.
The PR is now based on the current "develop" code state (2026.9.2-beta.1).
[Zip download](https://github.com/MoellerDi/homeassistant-stellantis-vehicles/archive/refs/heads/bugfix/logger-filter-accumulation.zip) for easy testing.

I will leave this PR in draft mode while the tests are still running.

chmtc94:
@MoellerDi 

I also needed to review my personal anonymizer script to make it more robust especially following your modifications for stored config, http headers and responses. My lazy initial script to extract json data from printed lines did not work any more...😉

I will post the updated version in corresponding discussion #562, just in case...

As already pointed, I however still wonder if it is a good thing to anonymize within the integration because we will definitively lost original values once masked, with no way to retrieve the original values if, for whatever reason, we would need them... Using an external script, we can keep all original logs, and the masked ones when needed.

I will however try your PR and give feedback asap😉

Again thanks a lot for your great support.👍

MoellerDi:
@chmtc94 

>I however still wonder if it is a good thing to anonymize within the integration because we will definitively lost original values once masked, with no way to retrieve the original values if, for whatever reason, we would need them...

Could you elaborate on that? For instance, I’m wondering why the access or refresh token would still be needed at a later stage. Do you have all fields in mind, or just specific ones?

In my opinion, masking serves simply to prevent users from exposing security-sensitive data, such as when creating a ticket here. Not everyone knows how to anonymize logs themselves beforehand.

chmtc94:
@MoellerDi 
> Could you elaborate on that? For instance, I’m wondering why the access or refresh token would still be needed at a later stage. Do you have all fields in mind, or just specific ones?

As I mentioned, for one reason or another—I couldn't say exactly which:
- it may be necessary to verify that the exposed sensitive values ​​(customer_id, VIN, email, password, secret code, etc.) actually match those used by the integration. 
- This obviously does not apply to tokens, where simply tracking usage—and using identical masked values for identical original values—should suffice. 
- Unless, perhaps, logs need to be shared with Stellantis (I’m probably dreaming a bit (a lot) here 😉), in which case the original logs would be required. 
 
My feeling is purely conservative, and  i agree probably useless in our case : always keeping original info, just in case....😉

> in my opinion, masking serves simply to prevent users from exposing security-sensitive data, such as when creating a ticket here. Not everyone knows how to anonymize logs themselves beforehand.

I however agree that for general user or general usage it would be nice to be able to share a log without having to anonymize it, and this function is quite a good one for them.

MoellerDi:
Rebased onto latest develop (2026.9.2-beta.2)

MoellerDi:
@chmtc94 How are your tests going? The code has been running without any issues on my instance since the PR was created. I am confident that this PR should resolve the performance and memory issues described in #414; however, since I never experienced these problems myself, I cannot personally verify this.

MoellerDi:
Rebased to 2026.9.3 codebase and consolidated the commits into two to keep the release notes concise.

I am confident that this PR should resolve the performance and memory issues described in https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/issues/414; however, since I never experienced these problems myself, I cannot personally verify this.

PR is IMHO ready for review.

chmtc94:
> @chmtc94 How are your tests going? The code has been running without any issues on my instance since the PR was created. I am confident that this PR should resolve the performance and memory issues described in #414; however, since I never experienced these problems myself, I cannot personally verify this.

Hi @MoellerDi 

Sorry for the delay. I'm not at home and it's a bit difficult for testing
I however just download your PR to my personal config.

No issue at startup.

I however quickly noticed some not hidden values for keys :
- client_id
- client_secret
- basic_token
- coordinates

May be also some not yet seen from the list in my anonymizer:

[secrets_values_example.json](https://github.com/user-attachments/files/32310038/secrets_values_example.json)

I will leave the test running to see if the #414 issue (excessive CPU usage) appear or not in some days...

EDIT : FYI , the PR runs with  Stellantis Vehicle  2026.9.3. 

EDIT2 : Many Update_data calls have a duration greater than 2 sec. I will perhaps need to reuse my log decorators to check which calls take more CPU time.

[log_durations.py](https://github.com/user-attachments/files/32311553/log_durations.py)

MoellerDi:
@chmtc94 

Thanks for the feedback and the tests. I’ve just marked this PR as a draft again to run some further tests.
`client_id`, `client_secret`, and `basic_token` (calculated from `client_id` + `client_secret` in `set_mobile_app`) are publicly visible in this repository (in `config.json`) anyway, so masking them isn't necessary.

I’ll take care of masking the coordinates this evening and push a fix.

chmtc94:
@MoellerDi 

I did not realize that client_id and client_secret was public known values available in configs.json. That look strange for me 🤔

> I’ll take care of masking the coordinates this evening and push a fix.

As present in my given list of handled keys (secret_values_example.json), pageToken values (not yet seen in log...) also need maybe to be anonymized 😉

PS : My personal anonymizer was a little blind but conservative and do not use any memory nor CPU ressource from the HA VM 😉

EDIT : Some potential secret values not masked in oauth token refresh url (I replaced potential sensitive values by <NOT_MASKED> in following log extraction. Again may be blind but conservative 😉) 

```
2026-09-17 07:59:12.677 DEBUG (MainThread) [custom_components.stellantis_vehicles.stellantis] HTTP exchange | url=https://idpcvs.driveds.com/am/oauth2/access_token?grant_type=refresh_token&refresh_token=<NOT_MASKED> headers={'Content-Type': 'application/x-www-form-urlencoded', 'Authorization': 'Basic <NOT_MASKED>'} | response={'access_token': '61c9c###', 'refresh_token': '742e3###', 'scope': 'email openid profile', 'id_token': '<NOT_MASKED>', 'token_type': 'Bearer', 'expires_in': 3599}
```

chmtc94:
@MoellerDi 

Below the log for Update_data durations since 15/09/2026 04H40 to compare with actual durations after this PR #593 update yesterday evening 16/09/2026 23H20

[2026-09-15-04H40-2026-09-17-10H00-update_data_durations.log](https://github.com/user-attachments/files/32327195/2026-09-15-04H40-2026-09-17-10H00-update_data_durations.log)

No signative differences observed until now before and after the update...Need however to wait more...

I also join the two error sequences that correspond to the 60 sec durations we can see in the previous log
(before and after the updating with the PR)

[2026-09-16-00H12-TIME-OUT-ERRORS.log](https://github.com/user-attachments/files/32327339/2026-09-16-00H12-TIME-OUT-ERRORS.log)

[2026-09-17-07H35-TIME-OUT-ERROR.log](https://github.com/user-attachments/files/32327340/2026-09-17-07H35-TIME-OUT-ERROR.log)

MoellerDi:
@chmtc94 

Thanks again for the detailed log excerpt, that made this easy to track down. Pushed a fix for the coordinates masking gap, and while going through the excerpt I found two more gaps in the same spot: `id_token` in the OAuth token-refresh response was never masked (only the initial code exchange registered it, not the periodic refresh), and the old `refresh_token` showed up unmasked in the refresh request's URL - it gets rotated out of the masked set right before that same log line is written with the old value still in the query string. Both are fixed now.

`pageToken` is just a pagination cursor for the trips endpoint (`get_vehicle_last_trip`), echoed back from the API's own `_links.last.href` - not a credential, so I don't think it needs masking, but let me know if you'd like it added anyway.

`client_id`/`client_secret`/`basic_token` stay unmasked on purpose, as discussed.

Still in draft while testing continues - thanks for sticking with it.

EDIT: [Zip download](https://github.com/MoellerDi/homeassistant-stellantis-vehicles/archive/refs/heads/bugfix/logger-filter-accumulation.zip) for easy testing.

EDIT2:
>I also join the two error sequences that correspond to the 60 sec durations

one time-out and the other an internal server error at Stellantis, both have been handled well.

chmtc94:
@MoellerDi 

Since I didn't know exactly—and still don't ☹️—which log information absolutely required anonymization versus what didn't, my custom anonymizer is designed to mask anything that *might* need masking. This includes all tokens, so-called "secret" keys, and all vehicle and client IDs and passwords. I didn't worry about whether masking was strictly necessary in every case; my only rule was: if it's suspicious, mask it. It’s a bit of a lazy approach, but it errs on the side of caution. The list of keys masked by my anonymizer is found in the file mentioned earlier (`secret_values_examples.json`); this includes, among others, the `id_token` and `pageToken` keys previously discussed.
My anonymizer allows you to adjust the anonymization process to your needs by modifying that file—adding or removing keys as required. (That is, provided you’ve kept the original, unmasked log file 😉).

Consequently, I don't have a firm opinion on the `pageToken` key, except to say that if it allows direct access to trip data, masking it might well be absolutely necessary...

andreadegiovine:
pageToken without oauth token dont work, but we can mask it like other stellantis customer data, client_id client_secret and basic_token need to be visible (anyway are already visible on repository) for debug session on users issues

MoellerDi:
@chmtc94 

`pageToken` is masked now too, and thanks @andreadegiovine for confirming `client_id`/`client_secret`/`basic_token` can stay visible.

The masking details aside, the CPU/memory behavior from #414 is the actual point of this PR, and I still can't reproduce it myself, so your feedback on that is what matters most here - whether it keeps happening or not. Please keep me posted either way once you've had a few more days on it.

MoellerDi:
Found `pageToken` leaking in another spot too - the vehicle list response (`get_user_vehicles`) echoes it back in `_links.last/self/first.href`, a different endpoint than the trips pagination I fixed earlier. Rather than patching every endpoint that happens to include one, I switched to masking `pageToken=...` wherever it shows up in a logged string, so this doesn't need chasing down endpoint by endpoint anymore. Pushed.

chmtc94:
I haven't updated to your latest PR yet, as I wanted to avoid restarting HA and continue monitoring CPU usage trends—just as I did with PR #414. 
So far, there has been no gradual drift in CPU usage, and the issue encountered in PR #414 has—as expected—completely disappeared. 
I therefore believe the PR can be fully approved; the latest adjustments (coordinates, id_token, and pageToken) give no apparent reason to change this assessment.

However, there is one minor point—seemingly unrelated to this PR, as it existed prior to its implementation—that I think warrants investigation: why do some calls to `get_vehicle_status` take over two (or even three or four) seconds, while others take less than 0.5 seconds? Analyzing and potentially optimizing certain sections of the code might be worth considering. 
To determine whether the cause lies with the integration or the Stellantis servers, I will add my `log_duration` decorators (https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/pull/593#issuecomment-5705029158) to a few methods called within `get_vehicle_status` in my next update, once your PR will be merged, just as I did for PR #414. 😉

Attached : 

1) System processor usage history from 14/09/2026 until now (Your PR was installed on 2026/09/16 23H30)

2026-09-14-2026-09-19-HA Processor usage

2) update_data() calls durations (extracted from lfull log)
[2026-09-16-08H07-2026-09-19-09H01-update_data_durations.log](https://github.com/user-attachments/files/32411356/2026-09-16-08H07-2026-09-19-09H01-update_data_durations.log)

3) Full log (twice anonymized from 16/09/2026 23H30 😉)
[2026-09-16-08H07-2026-09-19-09H01-home-assistant_anonymized.log](https://github.com/user-attachments/files/32411379/2026-09-16-08H07-2026-09-19-09H01-home-assistant_anonymized.log)

MoellerDi:
@chmtc94 Thanks for sticking with this and confirming #414 is gone - that's exactly what I was hoping to hear, especially since I could never reproduce it myself. Appreciate the graphs and logs too.

Agreed the `get_vehicle_status` duration variance is a separate matter, unrelated to this PR - let's track that in its own issue once you have your `log_duration` data.

@andreadegiovine PR is out of draft now, ready for review.

MoellerDi:
> However, there is one minor point—seemingly unrelated to this PR, as it existed prior to its implementation—that I think warrants investigation: why do some calls to `get_vehicle_status` take over two (or even three or four) seconds, while others take less than 0.5 seconds? Analyzing and potentially optimizing certain sections of the code might be worth considering.
> To determine whether the cause lies with the integration or the Stellantis servers, I will add my `log_duration` decorators ([#593 (comment)](https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/pull/593#issuecomment-5705029158)) to a few methods called within `get_vehicle_status` in my next update, once your PR will be merged, just as I did for PR #414. 😉

I just created another PR to add duration and CPU time logging to `log_call` 😃

chmtc94:
@MoellerDi @andreadegiovine 

Just making a test with trips, I just discover that the trip ids are not masked in get_last_trip responses.
As already said for others, I do not have a firm opinion on the necessity of masking these ids...?

[2026-09-18-14H32-2026-09-19-15H17-home-assistant_anonymized.log](https://github.com/user-attachments/files/32416248/2026-09-18-14H32-2026-09-19-15H17-home-assistant_anonymized.log)

PR #595, opened 2026-09-06
I/O cleanup in stellantis.py

MoellerDi:
Two tiny, independent robustness fixes in `stellantis.py`, one commit each.

## Commit 1: Drop redundant private `config_entries` save

`update_stored_config` called `self._hass.config_entries._async_schedule_save()` right after `async_update_entry(...)`. `async_update_entry` already schedules that save itself, so the extra call was redundant and poked at a private HA API. Removed the line; the config entry is still persisted through the public path.

## Commit 2: Download the vehicle picture via the HA client session with a timeout

`resize_and_save_picture` fetched the picture with `urlopen` and no timeout (default socket timeout is `None`), so a stalled CDN could hang the executor thread forever. The `ImageOps.pad(...)` decode also ran on the event loop.

- Fetch with `async_get_clientsession(...)` and `aiohttp.ClientTimeout(total=30)`, reading the bytes async.
- `resp.raise_for_status()` keeps the old "raise on HTTP error" behaviour; the caller in `get_vehicles` already handles that.
- Move the whole Pillow step into one `async_add_executor_job`, so nothing decodes on the event loop.
- Swap `from urllib.request import urlopen` for `from io import BytesIO` plus the `async_get_clientsession` import.

No config, entity, or translation changes. Both commits touch only the import block, `update_stored_config`, and `resize_and_save_picture`.

## Manual check

- [x] Reload the integration and confirm vehicle pictures still download to `www/stellantis_vehicles/<customer_id>/<vin>.png` (delete one to force a re-download).
- [x] Block the picture host and confirm setup no longer hangs, with a warning after ~30 s instead.
- [x] Toggle a stored-config value (e.g. remote commands) and restart HA to confirm it persisted.

### DIFF
--- a/custom_components/stellantis_vehicles/stellantis.py
+++ b/custom_components/stellantis_vehicles/stellantis.py
@@ -3,7 +3,7 @@ import aiohttp
 import base64
 from PIL import Image, ImageOps
 import os
-from urllib.request import urlopen
+from io import BytesIO
 from copy import deepcopy
 import paho.mqtt.client as mqtt
 import json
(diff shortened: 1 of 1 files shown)


chmtc94:
@MoellerDi 

A huge batch of PRs today to make this integration even more reliable and robust. You’re a true expert, and we’re really lucky to have you on this project. 😉👍🙏

Issue #597, opened 2026-09-08
STELLANTIS CONNECTION ISSUES 07/09/2026 AND 08/09/2026


chmtc94:
### Mark

DS

### Model

DS N°4

### Year

2025

### Engine

Hybrid

### Remote service

Connect Plus

### HomeAssistant version

Core 2026.8.3 Supervisor 2026.09.0 Operating System 18.2 Frontend 20260729.7

### Integration version

2026.9.1

### 1) 2026-09-07 10:45:55 - ERROR CVS error/user-vins - 50038 

   Automaticaly recovered at next update_interval

   Again appeared on: 

   - 2026-09-07 10:48:25 
   - 2026-09-07 11:31:11
   - 2026-09-07 15:55:13

   with apparent good automatic recovery.

### 2) 2026-09-07 19:13:46 - Cannot connect to host api.groupe-psa.com:443 ssl:default [DNS server returned answer with no data] 

   Not recovered until 2026-09-07 21:13:46 (EXACTLY TWO HOURS LATER ?)

### 3) 2026-09-08 00:21:52 - Same error

   Not recovered until 2026-09-08 01:21:52 (EXACTLY ONE HOUR LATER ?)

### 4) 2026-09-08 02:04:29 - Error communicating with Stellantis API

   This error was following a MQTT disconnection 

   Fully recovered 2026-09-08 02:10 after MQTT server reconnected (good handling)

### 5) 2026-09-08 07:14:02 ERROR : Cannot connect to host api.groupe-psa.com:443 ssl:default [DNS server returned answer with no data]

   NOT RECOVERED UNTIL NEXT HA RESTART 
 
### 6) 2026-09-08 07:50 =>  HA RESTART

No more connection errors...

### 7) 2026-09-08 08:10 =>  HA RESTART

_(Nota 1 :  Following sv_dashboard integration 0.6.0.beta.12 update)_
_(Nota 2:  Two commands sent via MQTT (from sv_dashbord thru Stellantis Vehicles) : Start conditionning and stop conditionning)_

No connection errors until 2026-09-08 08:35:16

### 8) 2026-09-08 08:35:16 - ERROR : Cannot connect to host api.groupe-psa.com:443 ssl:default [DNS server returned answer with no data]

No automatic recovery until HA RESTART

### 9) 2026-09-08 08:46:26 HA RESTART

Connection errors again disappeared

### 10) 2026-09-08 08:46 to 11:00  

[...]

PR #598, opened 2026-09-08
Add exponential backoff to the MQTT token refresh retry

MoellerDi:
## Summary

`scheduled_mqtt_token_refresh` reschedules itself a flat 60 seconds later on every `CommunicationError`, with no log line in that branch. When the Stellantis token backend has a sustained outage (an intermittent `401 Unauthorized - Cannot pass the security checks ...`, or `400 server_error: "Could not read token in CTS: Failed to read Token"`), the integration retries the token endpoint once a minute for as long as the outage lasts, and none of it is visible unless debug logging is on. Issue #597 shows roughly 18 such retries in 18 minutes.

This changes the retry to a capped backoff schedule with jitter, resets the counter on success, and logs each failed attempt at WARNING so the retry loop is visible without debug logging. The regular (non-error) refresh cadence, still anchored to the token expiry, is unchanged.

## Change

`const.py`:

- New `MQTT_TOKEN_RETRY_BACKOFF = (60, 120, 300, 600, 900)`: the retry delay in seconds after the 1st, 2nd, ... consecutive failure, capped at the last step (so a prolonged outage is retried every ~15 minutes instead of every minute).

`stellantis.py`, `scheduled_mqtt_token_refresh`:

- New instance counter `self._mqtt_token_retry` (init `0`).
- Success path (a completed `refresh_mqtt_token_request()`, or a call made while no refresh was due): reset the counter to `0`, then schedule from `get_next_run()` as before.
- `except CommunicationError`: increment the counter, take `MQTT_TOKEN_RETRY_BACKOFF[min(retry - 1, len - 1)]`, add up to 10% jitter (`random.uniform(0, delay * 0.1)`, so several config entries do not retry in lockstep), schedule that many seconds out, and log `MQTT token refresh failed (attempt N), next retry at ...` at WARNING.
- `except RateLimitException` and `except ConfigException`: unchanged behaviour, plus a counter reset to `0`.

## Behaviour

| | Before | After |
|---|---|---|
| Single transient failure | retried after 60 s, no log line (debug-only at the end) | retried after ~60 s, one WARNING with the attempt number and reason |
| Sustained token-backend outage | retried every 60 s for the whole outage | retried after ~60 s, 2 m, 5 m, 10 m, then every 15 m |
| Recovery | n/a (counter did not exist) | next success resets the counter; the following failure starts again at 60 s |
| Regular refresh (no error) | scheduled at token expiry minus 3 min (~12 min out) | unchanged |
| Rate limit / bad OTP config | 1 day / disable remote commands | unchanged (counter also reset) |

### DIFF
--- a/custom_components/stellantis_vehicles/const.py
+++ b/custom_components/stellantis_vehicles/const.py
@@ -131,6 +131,11 @@ UPDATE_INTERVAL = 60 # seconds
 # re-fetched to check whether the vehicle was unpaired.
 EMPTY_STATUS_LIMIT = 3
 
+# Backoff schedule (seconds) for the MQTT token refresh after a transient
+# Stellantis backend failure. Capped at the last step so a prolonged outage is
+# retried every ~15 min instead of once a minute.
+MQTT_TOKEN_RETRY_BACKOFF = (60, 120, 300, 600, 900)
+
 VEHICLE_TYPE_ELECTRIC = "Electric"
 VEHICLE_TYPE_HYBRID = "Hybrid"
 VEHICLE_TYPE_THERMIC = "Thermic"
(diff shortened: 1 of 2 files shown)


MoellerDi:
successfully tested, works - ready for review

PR #600, opened 2026-09-11
Retry once after a transient 401 before failing

MoellerDi:
A transient `401` right after an OAuth token rotation surfaces as `UpdateFailed("Error communicating with Stellantis API")`, even though the next poll usually succeeds on its own.

**What this PR change**

- `make_http_request(..., _retried=False)`: on a first-attempt `401` for a non-token-endpoint URL, call `refresh_oauth_token_request()` once, swap the fresh bearer into the `Authorization` header and retry. If it still fails, raise `CommunicationError("Stellantis rejected the access token (HTTP 401)")` instead of the raw server text.
  - `ConfigEntryAuthFailed` from a dead refresh token still propagates, so reauth triggers normally.
  - `refresh_oauth_token_request` is already `@rate_limit(6, 1800)`, so the retry can't storm the token endpoint.
- `base.py::_async_update_data`: `UpdateFailed` now points users at debug logging for the real reason, instead of a bare generic string.
- `make_http_request`: the `404`, `500`, `400`/`invalid_grant` and generic `50x` branches now fall back to a fixed, descriptive message when the response body doesn't match a known error shape, instead of raising/logging the literal string `"None"`.

**Behaviour**

| | Before | After |
|---|---|---|
| Transient 401 | one `UpdateFailed`, entities flap to `unavailable` for a cycle | silently refreshed and retried, no `UpdateFailed` |
| Persistent 401 | generic `UpdateFailed`, no pointer to the debug log | same generic `UpdateFailed`, now points at the debug log (raw server text was and remains debug-only) |

### DIFF
--- a/custom_components/stellantis_vehicles/base.py
+++ b/custom_components/stellantis_vehicles/base.py
@@ -86,7 +86,9 @@ class StellantisVehicleCoordinator(DataUpdateCoordinator):
             raise
         except Exception as err:
             _LOGGER.debug("Error communicating with Stellantis API: %s", err)
-            raise UpdateFailed("Error communicating with Stellantis API") from err
+            raise UpdateFailed(
+                "Error communicating with Stellantis API, enable debug logging for details"
+            ) from err
 
         if not new_data:
             # Keep the last known data instead of blanking every entity on a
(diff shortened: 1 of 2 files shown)


MoellerDi:
successfully tested, works - ready for review

PR #605, opened 2026-09-13
Make coordinator.data the actual data, and clean up the update cycle around it

MoellerDi:
## Summary

`StellantisVehicleCoordinator._async_update_data()` never returned a value, so `DataUpdateCoordinator` kept assigning `None` to `coordinator.data` on every refresh cycle, forever. Nothing depended on it directly since every internal read went through the private `self._data`/`self._sensors` attributes instead, but it meant:

- the `if self._coordinator.data is False:` check in `_handle_coordinator_update` (base.py) could never be true, dead since the very first commit of this repository.
- the framework's own data path was simply unused, so anything relying on the standard `DataUpdateCoordinator` contract (external tooling, future code, tests) would find `coordinator.data` empty even right after a successful poll.

This branch carries five commits:

1. `Make coordinator.data the single source of vehicle data`: every return path of `_async_update_data` now returns the vehicle status, `self._data` is removed, and the three places (base.py x2, diagnostics.py) that read the private `_data` attribute now read the public `.data` property instead. `after_async_update_data`/`send_abrp_data` take the freshly fetched payload as an explicit `new_data` parameter, since `self.data` is not updated by the framework until the whole update cycle returns; `self._sensors` (a different, independently maintained cache, see "Scope and non-goals") is still read as before. The dead `is False` check is removed.
2. `Guard send_abrp_data against a null lastPosition and extract it once`: `lastPosition` can be `null` in the API response, the device tracker already accounts for that (`_last_position`); `send_abrp_data` did not, and indexed into it four separate times. The position is now extracted once, with the same guard the device tracker uses.
3. `Split _async_update_data into named phases`: the method mixed six different concerns (interval bookkeeping, API fetch and error translation, empty-response handling, success bookkeeping, staleness check, apply and return) in 73 lines. It is now a 14-line list of calls to four small, single-purpose methods.
4. `Split after_async_update_data into three focused helpers`: same treatment for the three unrelated post-update actions (automatic charge-limit stop, ABRP sync, last-trip fetch on engine stop), which were nested up to four levels deep in one method.
5. `Remove unused variable and guard get_programs against missing preconditioning data`: found while reviewing the above. `get_programs()` indexed `preconditionning`/`airConditioning` without a default, so a status response missing that section raised `KeyError`. It now falls back to an empty dict at each step; an unused local variable is also removed.

## Change

`base.py`, `StellantisVehicleCoordinator`:

- `_async_update_data` is now an orchestrator calling `_apply_refresh_interval_override`, `_fetch_new_data`, `_handle_empty_status`, `_is_stale`, in that order, and finally `after_async_update_data(new_data)`. Each of those is a direct extraction of the equivalent block from the old method. `_handle_empty_status`'s short-gap branch returns `self.data or {}`, so `coordinator.data` keeps the same "always at least an empty dict" guarantee the old `self._data` attribute had, even on a vehicle's very first poll.
- `after_async_update_data(new_data)` now calls `_auto_stop_charge_at_limit()`, `_sync_abrp_if_enabled(new_data)`, `_fetch_last_trip_on_engine_stop(new_data)`. Same relationship: each helper is the old nested block, turned into guard clauses instead of nested `if`/`elif`.
- `send_abrp_data(new_data)` builds `coordinates`/`heading` once at the top from `new_data.get("lastPosition")` (guarded with the same `isinstance(..., dict)` check as `_last_position`), instead of repeating the lookup chain at each of the three places that used it.
- `get_programs()` reads `air_conditioning = (self.data.get("preconditionning") or {}).get("airConditioning") or {}` instead of indexing directly.
- `get_value_from_map`, `_last_position`: read `self._coordinator.data` instead of `self._coordinator._data`.
- `_handle_coordinator_update`: the dead `if self._coordinator.data is False: return` is removed.

`diagnostics.py`:

- `_coordinator_diagnostics` reads `coordinator.data` instead of `coordinator._data` (the `# noqa: SLF001` this needed is gone with it).

No change to `self._sensors`, its population logic, or any of the entity platform files.

## Behaviour

| | Before | After |
|---|---|---|
| Any successful poll | `coordinator.data` stays `None` forever | `coordinator.data` holds the vehicle status |
| Single empty status response (short gap) | `coordinator.data` would have been reset to `None` if it were ever read | `coordinator.data` keeps the last known status |
| `send_abrp_data` when `lastPosition` is `null` | `AttributeError` | ABRP payload sent with `lat`/`lon`/`heading`/`elevation` left at their defaults |
| `get_programs()` when the status has no `preconditionning` section | `KeyError` | returns the default (all slots disabled) programs |

Everything else (entity states, sent commands, polling cadence, log lines) is unchanged for a vehicle that responds normally.

## Scope and non-goals

- `self._sensors` (the cache of already-converted, per-entity values, some of which are user-configured switch/number/text settings that have no equivalent in the API payload) is a different data source from `coordinator.data` (the raw API payload) and is left exactly as it was. Folding it into `coordinator.data` too was considered and rejected: several of its values do not exist in the API response at all, and the ones that do would need the entities' path-resolution logic (`get_value_from_map`) duplicated inside the coordinator.
- No config schema, entity, or service changes.
- No behaviour change for a vehicle that responds normally; every change here is either a dormant-bug fix (see the table above) or a pure readability refactor.

## Manual verification on my setup

 - [x] Reload the integration. Entities populate as before; `coordinator.data` diagnostics field (`raw_status` in the diagnostics download) is populated, not empty.
 - [x] Enable debug logging, let a few poll cycles run, confirm the log lines (`Empty vehicle status response...`, `API did not return updated vehicle data...`, engine-stop trip fetch) still appear at the expected moments.
 - [x] With ABRP sync enabled and a valid 36-character token, confirm the ABRP payload still carries the correct lat/lon/heading after a normal poll.
 - [x] With the charge-limit switch and a limit configured, start a charge and let the battery reach the limit; confirm the delayed charge-stop command is still sent exactly once per charge session.
 - [x] Trigger a `ThermalPrecond`/preconditioning program command and confirm the four program slots round-trip correctly.


MoellerDi:
@andreadegiovine May I ask you to pre-review this PR and let me know if it goes in the right direction or if you prefer something else?

andreadegiovine:
If it work and dont break any features is the right way, thanks for your big support!

MoellerDi:
Thank you for your kind feedback. I will carry out the final tests to get everything ready to go.

MoellerDi:
@andreadegiovine: all tests successfully completed, ready for review.

PR #613, opened 2026-09-15
Add capped backoff and logging to the OAuth token refresh retry

kZou13:
## Summary

When the scheduled OAuth refresh fails with a `CommunicationError`, `scheduled_oauth_token_refresh` retries a flat 5 minutes later, with no log line in that branch, for as long as the outage lasts. `scheduled_mqtt_token_refresh` got a capped backoff with a WARNING per attempt in #598; this does the same for the OAuth side.

The normal cadence is unchanged: the refresh still runs 5 minutes before expiry, so there is no additional token rotation. (An earlier version of this PR moved that margin to 15 minutes; the discussion below showed that would only add rotations, since the refresh token outlives the access token and the observed failures were lost responses after a rotation. That part is dropped.)

## Change

`const.py`: `OAUTH_TOKEN_RETRY_BACKOFF = (60, 120, 300, 600, 900)`, same schedule and semantics as `MQTT_TOKEN_RETRY_BACKOFF`.

`stellantis.py`, `scheduled_oauth_token_refresh`: a counter `self._oauth_token_retry`, reset to 0 on a completed refresh; on `CommunicationError`, take `OAUTH_TOKEN_RETRY_BACKOFF[min(retry - 1, len - 1)]` with up to 10 % jitter, schedule that far out, and log the attempt at WARNING, mirroring #598.

## Behaviour

| | Before | After |
|---|---|---|
| Sustained refresh outage | retry every 5 min, silent | 1, 2, 5, 10 then 15 min between attempts, each logged at WARNING |
| Normal refresh | at expiry − 5 min | unchanged |

## Testing

Applied against `develop` and syntax-checked; the backoff form has not run for a week here yet. Reconfiguration, commands and sensors are unaffected. I have not tested a fresh installation on this branch.

Independent of #612, which touches the exception handlers of the same function; if one lands first I will rebase the other.


MoellerDi:
EDIT: Since I didn't want to be rude...

Good point, by the way... but please note that `OAUTH_REFRESH_TOKEN_QUERY_PARAMS` (`"grant_type": "refresh_token"` or `{#oauth|refresh_token#}`) is used in `refresh_oauth_token_request`. The logic mirrors that of MQTT tokens, where the refresh token is valid for about 7 days (and is renewed every 3 days as a precaution [MQTT_REFRESH_TOKEN_TTL]). It is not yet clear how long the OAuth refresh token remains valid.

EDIT: Regular API calls use `CAR_API_HEADERS` (`Bearer {#oauth|access_token#}`).

kZou13:
Thanks — and thanks for the edit, that is the distinction that matters. Going back over the two failures with it, I see two readings of the same log, and I cannot separate them from what I have:

**A — the refresh token dies with the access token.** The retry landed about a minute after `expires_in` and was refused. Consistent with the refresh token being scoped to the access token's validity, or both expiring together. Under A, this PR fixes the actual failure.

**B — a lost response after rotation.** The first attempt (at `expires_in − 5`) reached the server, which rotated the refresh token, but the response never came back — the logged error had an *empty* message, which is what a dropped response looks like, not a refusal. The client kept the old, now-invalid token, and the retry got `invalid_grant` regardless of timing. Under B an earlier margin does not help that case (it still helps genuine failures where nothing reached the server).

For what it is worth, `refresh_oauth_token_request` stores the refresh token from every response, so in normal operation it is rotated every cycle and never ages; the only time it does is when the chain stops, which is #612's subject.

One test separates A from B without stopping Home Assistant: disable the config entry ~20 min before `expires_in` (`async_shutdown` cancels the refresh timers; 20 rather than 10 so it also works with this PR's 15-minute margin), re-enable it ~45 min after. On load, `scheduled_oauth_token_refresh` sees `now > get_next_run()` and refreshes with the stored token. Success → B, and a >60 min outage costs nothing. `invalid_grant` → A. Do you have a test account or environment to run that on? If not, I can run it here and report back — worst case is one reauth.

Separately, since the server tells us the access token lifetime on every refresh, I would rather not hard-code the margin against it: I will make it proportional — `min(OAUTH_TOKEN_REFRESH_MARGIN, expires_in * 0.25)`, i.e. 15 min for today's 3600 s token, and still sane if a shorter one is ever issued. That removes this PR's own assumption; to be clear, it says nothing about the refresh token question above. I will push it once you have had a look at the A/B point, so the diff does not move under you.

MoellerDi:
In my opinion, we can rule out the scenario where "the refresh token becomes invalid along with the access token," because if that were the case, the retry logic for 401 errors wouldn't work. I just tested this on my system by completely disabling `scheduled_oauth_token_refresh`, commenting it out as follows (requires an HA restart):
```
async def scheduled_tokens_refresh(self):
    self.reset_scheduled_tokens()
    #await self.scheduled_oauth_token_refresh()
    await self.scheduled_mqtt_token_refresh()
```
You can see that a valid set of OAuth tokens is maintained (both access and refresh tokens, since the backend response includes both). However, I am concerned that triggering so many 401 errors might attract unwanted attention from Stellantis; therefore, it is better to refresh the access token proactively and in a timely manner.

MoellerDi:
I also noticed a while ago that I can shut down my development or test environment for days (if not weeks) without normally having to re-authenticate. That is why I am fairly certain that it involves a long-lived OAuth refresh token.

kZou13:
That settles it, thank you. Disabling the scheduler is a cleaner experiment than the one I proposed, and it rules A out: the refresh token outlives the access token. So the two failures I saw were B, a lost response after rotation, and nothing client-side can recover from that except reauth, which is what #612 makes immediate.

That changes what this PR is for, and I have updated the description accordingly. It is no longer "one transient error costs the session". It is: refresh proactively and early enough that a poll never has to fall back on a `401`, with a capped, logged retry schedule instead of a flat 5-minute loop. Your point about not attracting attention with `401` storms is exactly the argument.

On the margin, I intend to make it proportional to `expires_in` (`min(OAUTH_TOKEN_REFRESH_MARGIN, expires_in * 0.25)`, so 15 min for today's 3600 s token). The server states the lifetime on every refresh, so I would rather derive the margin from it than assume it. Do you see a drawback I have missed? I am preparing that change and will push it once you have answered, so the diff stays put while you read.

MoellerDi:
I am not sure if we need to take an aggressive approach, given that the retry logic for the 401 error kicks in anyway once Stellantis stops accepting our OAuth access token. If a response were indeed lost following a rotation, frequent active rotation would only increase that risk.

kZou13:
You are right: with A ruled out, refreshing earlier only adds rotations, and each rotation is a chance to lose a response. I have dropped the margin change (and the proportional idea) and reduced the PR to the failure path only, a capped, logged backoff on `CommunicationError` in place of the flat 5-minute retry, same shape as #598. The normal cadence is untouched. Title and description updated to match.

Issue #614, opened 2026-09-15
Battery guard from #463 reads a stale cached autonomy and freezes the battery sensor at 0%


kZou13:
### Mark

Peugeot

### Model

3008

### Year

2020

### Engine

Hybrid

### Remote service

E-remote control

### HomeAssistant version

2026.9.2

### Integration version

2026.9.2

### What happened?

The battery guard added for #463 reads the autonomy from the retained sensor cache, not from the payload the battery value came from. When the two disagree, a fresh 100 % is rewritten to 0 % and the sensor then freezes there.

**What I saw**

29 Aug, 15:20:51, end of a charge: the cloud reported ~100 % and `sensor.<vin>_battery` showed **0 %**, stuck for 10 h 28 min. The car was plugged in the whole time.

**Why, reading `base.py` `get_value`**

```python
if key == 'battery':
    value = int(value)
    autonomy = self._coordinator._sensors.get("autonomy")
    if value > 90 and autonomy is not None and autonomy == 0:
        value = 0
```

Three things line up:

1. `self._coordinator._sensors` is a retained cache — a few lines above, `if value is not None or key not in self._coordinator._sensors` keeps the previous value whenever a field is **absent** from the current payload. `autonomy` is absent while the car sleeps, which is the whole of a charge, so the cache still holds the `0` from the previous day.
2. The guard reads that cache, not the payload. `battery` arrives fresh at 100, `autonomy` reads the stale `0`, and the guard rewrites 100 → 0.
3. The raw value was already stored in the cache *before* the guard ran, so `value_was_updated()` compares `100 == 100`, concludes nothing changed, and the entity is never written again. The `0` is frozen until the raw value itself moves.

Point 3 is what turns a one-cycle glitch into a stuck sensor.

**How often the guard actually helps here**

Over 1 180 h of recorder history it never caught a real API glitch on this vehicle — including 14 consecutive hours at "100 % / 0 km" on 27-28 Aug, where it did *not* fire because the freshly reported autonomy was not 0. The only time it fired, it was wrong.

**Suggested fix**

Read the autonomy from the current payload, using its own value map (`const.py`: `["energies", {"type": "Electric"}, "autonomy"]`):

```python
autonomy = self.get_value_from_map(["energies", {"type": "Electric"}, "autonomy"])
```

[...]

Issue #535, opened 2026-08-18
Bug: Battery level reported as 100% while residual kWh and capacity reflect actual charge


RuiAlexSaint:
### Mark

Peugeot

### Model

e2008

### Year

2025

### Engine

Electric

### Remote service

E-remote control

### HomeAssistant version

Core 2026.8.2 / Supervisor 2026.07.5

### Integration version

2026.7.2

### What happened?

First of all, thank you for maintaining this great integration!

I would like to report a recurring data discrepancy issue regarding the battery level (battery sensor) on certain EV models (such as the e-CMP platform).

The Issue
The cloud backend frequently returns the raw battery percentage (level) stuck at 100%, even when the vehicle is partially discharged. However, the energy telemetry fields (battery_residual and battery_capacity) are correctly updated and reflect the real energy state of the battery. For instance, the car might report a residual energy of 12.15 kWh out of 52.70 kWh (approx. 23%), but the main battery percentage sensor still displays 100%.

Suggested Solution
Since all necessary energy metrics (battery_residual and battery_capacity) are already fully collected and processed within the base.py (inside the get_value method for the 'battery' key), we can implement a reliable fallback/correction mechanism directly at that level.

If the calculated percentage from (residual / capacity) * 100 is lower than the raw level reported by the cloud API, the integration can safely override and assume the calculated value.

### Log output

```shell

```

Jordan87:
You should rise this topic to Stellantis.
I provided all the data and graphs from my side and according to support they are working on it.
More the tickets, more probably that root cause will be fixed.

chmtc94:
> You should rise this topic to Stellantis. I provided all the data and graphs from my side and according to support they are working on it. More the tickets, more probably that root cause will be fixed.

hI @Jordan87 @RuiAlexSaint 

What is the link to support you suggest to rise the issue to ?

I also encountered a similar issue with battery level when the level reported in API data suddenly drop to 0% with no apparent reason (autonomy value remaining at the same value) until a wake-up restore the right level value.

(NB : As highlighted in the screenshot and the log: I change the update_interval before the trip I made before the charging session. The issue happened before the charging session and after the car was plugged)

<img width="834" height="1210" alt="Image" src="https://github.com/user-attachments/assets/764b9fdb-2fab-4056-9bd3-7678dea23e83" />

and associated log file: 

[...]

Issue #463, opened 2026-04-23
Battery 100% on HA even if in reality it is a 0%


Destroyer061090:
### Mark

Peugeot

### Model

508SW

### Year

2020

### Engine

Hybrid

### Remote service

E-remote control

### HomeAssistant version

2026.4.3

### Integration version

 2026.3.6

### What happened?

<img width="577" height="501" alt="Image" src="https://github.com/user-attachments/assets/d4a94486-4159-4781-a0ed-7a11c92bcfb6" />

I see 100% of the battery but in reality I have 0% 

### Log output

```shell

```

sod75:
I have this on my 3008hybrid as well, when the battery is drained Battery autonomy is zero but battery percentage says suddenly 100% or sometimes 67%.
In my opinion this is the value coming from the API, not a HA or integration bug.

Destroyer061090:
In my case I have battery autonomy range to 0km and battery 100%... from my point of view a change in the custom integration whould help... something like if battery autonomy range is 0 then even battery should be 0.... because make no sense to have 100% battery but not autonomy range.... @andreadegiovine  what you think?

<img width="328" height="212" alt="Image" src="https://github.com/user-attachments/assets/e66cfbf5-ddcd-4e79-8460-beb65b341528" />

Destroyer061090:
Additionally i tryed to modify base.py adding this fix which seems working...

inside "def get_value(self, value_map):" just before "return value"

```
# --- FIX_BUG API STELLANTIS ---
        # Caso noto: autonomia = 0 km ma batteria = 100%
        if (
            key == "battery"
            and value is not None
            and value > 90
            and self._coordinator._sensors.get("autonomy") == 0
        ):
            _LOGGER.warning(
                "Stellantis API bug detected: autonomy=0 km, battery=%s%%. Forcing battery to 0%%.",
                value,
            )
            value = 0
            self._coordinator._sensors[key] = value
        # --- FINE FIX ---
```

Jordan87:
I also have that issue with my 308 PlugInHybrid that sometimes suddenly when its charging it jumps to 100% and during next wakeup it drops to normal level.
This is clearly issue on server/car side.

Destroyer061090:
Still present

Destroyer061090:
still present. @andreadegiovine have you evaluate the proposal I wrote? I test it and and least solve the wrong value on HA

PR #615, opened 2026-09-15
Evaluate the battery guard on the current payload instead of the cached autonomy

kZou13:
## Summary

The battery guard added for #463 reads `autonomy` from the retained sensor cache rather than from the payload the battery value came from. When the two disagree, a fresh `100` is rewritten to `0` and the sensor then freezes there. Full reasoning, measurements and history in #614.

In short:

- `self._coordinator._sensors` keeps the previous value whenever a field is absent from the payload — the case for the whole of a charge, while the car sleeps — so the cached `autonomy` can still hold yesterday's `0`.
- The guard reads that cache, so a freshly reported `battery = 100` is turned into `0`.
- The raw value was already stored in the cache before the guard ran, so `value_was_updated()` compares `100 == 100`, and the entity is never written again. The `0` is frozen until the raw value itself changes.

Observed here on 29 Aug: `0 %` held for 10 h 28 min at the end of a charge, vehicle plugged in, while the energy fields kept reporting a full pack.

## Change

`base.py`, `get_value`: resolve the autonomy with its own value map — `self.get_value_from_map(["energies", {"type": "Electric"}, "autonomy"])` — instead of `self._coordinator._sensors.get("autonomy")`, so both values come from the same payload.

## Behaviour

| payload | Before | After |
|---|---|---|
| `battery = 100`, `autonomy = 0` (the API bug of #463, still seen in #535) | battery → 0 | battery → 0 (unchanged) |
| `battery = 100`, no autonomy field, cached autonomy = 0 | battery → 0, then frozen | battery = 100 |
| `battery = 100`, `autonomy = 32` | 100 | 100 |

The guard is not weakened: every payload that actually carries the contradiction still triggers it. Only the case where the autonomy is simply *missing* stops being treated as "autonomy is zero".

## Testing

The guard has been disabled locally since 30 Aug without regression, which is how the freeze was identified, but that is not the same change as this one. **This exact form has been applied against `develop` and syntax-checked; it has not yet run for a week here.** I will report back after a full week if you would rather wait for that before merging.

Reconfiguration, commands and other sensors are unaffected — the change is a single expression inside `get_value`. I have not tested a fresh installation on this branch.

## Note

Opened alongside #614 rather than instead of it: this is your fix from #463 and I did not want to touch it without laying out the reasoning first. Happy to close this PR if you prefer a different approach — for instance testing the *freshness* of the cached autonomy rather than reading the payload directly.

Independent of #612 and #613, which touch `stellantis.py` only.

Refs #614


chmtc94:
@kZou13 

Could you have a look to this comment https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/issues/614#issuecomment-5705377245

kZou13:
Answered in #614: your log shows the server itself returning `level: 0.0` with a fresh `createdAt`, which is a different mechanism from the guard this PR changes. Details there.

Issue #77, opened 2025-03-15
PL language update and few questions


Jordan87:
### Mark

Peugeot

### Model

308 PlugInHybrid

### Year

2025

### Engine

Hybrid

### Remote service

E-remote

### What happened?

First I want to say big Thank You for your addon. 

I was using psacc for over a year but it have some strange problems quite often (status is updating if I drive or do it from MyPeugeot app but not from the psacc addon). At least for now your addon is working perfectly fine.

Because I have never used github to pull/fork anything im attaching corrected translation here:)
[pl.json](https://github.com/user-attachments/files/19261105/pl.json)

How works "Refresh data interval" parameter? Is it about how often it reads from server i.e. if I drive? (btw its not in translation file, in polish it should be "Częstotliwość odświeżania danych" ;) )

Maybe you can add not in readme that sensors name can be different for diffrent languages then english (I spend around 15min to figure out why its not accepting my automation for refreshing status during charging :) )

Im also thinking to reconfigure that automation not to trigger it always with configured interval (and check condition if its charging) but to count i.e. configured interval from the moment when I start charging (currently it may happend that wakeup command will be send even several seconds from last update when charging was started.

Do you have any plans to add in a distant future, history of trips and charges like in psacc? It is not necessery for me but was a nice addon :)

Is there any gain of using ABRP for plugin hybrid cars? I tried to find some informations but it seems that it is useful only for full EVs.

Big thanks again for your work and time. Small gratitude was sent from me ;)

### Log output

```shell

```

andreadegiovine:
@Jordan87 the next update will include your pl translations updated.

> How works "Refresh data interval" parameter? Is it about how often it reads from server i.e. if I drive? (btw its not in translation file, in polish it should be "Częstotliwość odświeżania danych" ;) )

Is the interval to get data from server, in en.json translation is present, add it to your language.

> Maybe you can add not in readme that sensors name can be different for diffrent languages then english (I spend around 15min to figure out why its not accepting my automation for refreshing status during charging :) )

Im Italian :) i try to write this integration in english only for the community, Which sensors are not understood?

> Im also thinking to reconfigure that automation not to trigger it always with configured interval (and check condition if its charging) but to count i.e. configured interval from the moment when I start charging (currently it may happend that wakeup command will be send even several seconds from last update when charging was started.

[...]

PR #616, opened 2026-09-16
Add native trip history transport API

CaneTLOTW:
## Summary

Adds a small native transport method for historical trips:

```python
await stellantis.get_vehicle_trips(vehicle, since=None, page_token=None)
```

The method follows the existing `get_vehicle_last_trip()` style and reuses the integration's existing authenticated client/session, headers and `/trips` endpoint.

## Motivation

A downstream Home Assistant companion project, [CaneTLOTW/sv_dashboard](https://github.com/CaneTLOTW/sv_dashboard), needs historical `/trips` data. Without a native upstream method it currently has to reproduce the request wiring against internal transport details.

This PR intentionally keeps the upstream change minimal: expose one page of raw historical trip data and leave pagination/aggregation/storage/UI to the caller.

## Changes

- add `get_vehicle_trips(vehicle, since=None, page_token=None)` to `StellantisVehicles`;
- always request `distance=0.1-`;
- optionally pass `timestamps=<since>/`;
- optionally pass `pageToken`;
- return the raw Stellantis response unchanged;
- reuse the existing `make_http_request()` path and shared authentication/session;
- remove the old commented-out `get_vehicle_trips` stub.

## Compatibility / scope

- `get_vehicle_last_trip()` is unchanged;
- no new Home Assistant service/entity/coordinator state;
- no additional OAuth flow or aiohttp session;
- no pagination loop or history aggregation in upstream;
- no version change.

This keeps the change aligned with the existing transport helpers while providing a stable integration point for consumers that need trip history.

Related earlier discussion: #77.

## Validation

- `python -m compileall custom_components/stellantis_vehicles` — PASS
- focused transport test for vehicle/since/page token/raw response — PASS

HACS/hassfest checks can run on this draft PR.


CaneTLOTW:
## Related context

- #77 — earlier upstream discussion where trip history was intentionally limited to the last trip.
- [Discussion #525](https://github.com/andreadegiovine/homeassistant-stellantis-vehicles/discussions/525) — community / DS4 Hybrid testing that helped evolve the original e-C3-specific dashboard into the generic Stellantis Vehicle Dashboard.
- [SV Dashboard DS4 Hybrid / French external test](https://github.com/CaneTLOTW/sv_dashboard/issues/2) — current successor-project test thread.

This PR does not move dashboard/history behavior upstream. It only provides the small native transport method required to avoid coupling downstream consumers to private request internals.

chmtc94:
@andreadegiovine @MoellerDi 

Having a look at  the Stellantis API developer documentation concerning trips (https://developer.groupe-psa.io/webapi/b2c/api-reference/references/#tag/Trips), I wonder if using the indexRange parameter would not be a better way to access the last trip data, rather than the timestamps parameter that do not allow to get old last trips when the vehicle was not used for a while.  🤔 Duration and/or Distance parameters can still filter 0 kms trips (start/stop engine without moving the vehicle).

andreadegiovine:
Currently the integration request last_trip on engine shutdown, then if you keep your vehicle parked there are no new trips before the next shutdown (then a new trip), yes we can use this parameter but is not usefull

custom_components/stellantis_vehicles/stellantis.py
```diff
@@ -733,6 +733,26 @@ async def get_vehicle_last_trip(self, vehicle, page_token=None):
                 return await self.get_vehicle_last_trip(vehicle, next_page_token)
         return vehicle_trips_request
 
+    @log_call
+    async def get_vehicle_trips(self, vehicle, since=None, page_token=None):
+        """Get one page of historical trips from Stellantis."""
```
andreadegiovine: Add a comment with url of third party component, to prevent future edit

chmtc94:
@andreadegiovine 

> Currently the integration request last_trip on engine shutdown,

I planned to use it at startup for the first get last trip not following a engine stop event

CaneTLOTW:
@chmtc94 @andreadegiovine thanks — the new replies clarify the two different use cases.

@andreadegiovine is right for the **current integration flow**: `get_vehicle_last_trip()` is called when the engine changes to `Stop`, so the trip has just ended and the current one-day timestamp window is normally sufficient there.

@chmtc94 is describing a different use case: an **initial last-trip fetch at startup**. In that case, if the vehicle has been parked for more than 24 hours, the current timestamp filter can return no trip even though an older last trip exists.

For SV Dashboard this is separate from PR #616:

- SV Dashboard does **not** use upstream `get_vehicle_last_trip()` for canonical history.
- PR #616 only asks for the raw authenticated `get_vehicle_trips(vehicle, since=None, page_token=None)` transport.
- SV Dashboard keeps pagination, overlap, deduplication, persistence and gap recovery itself.

We also tested `indexRange` read-only against the production API.

Observed behavior:

- `indexRange=0-0` = oldest trip;
- global `total - 1` = newest trip;
- `pageSize=1` returns the same first trip as index 0 and preserves the global `total`;
- following `_links.last` returns the same newest trip as the explicit final global index;
- when `indexRange` is used, the response `total` becomes the size of the requested range (`1`, `5`, ...), so the global final index has to be obtained separately first.

Because of that, I would **not switch the history implementation to `indexRange`**.

For history synchronization, the existing timestamp + page-token/HAL pagination approach is the better fit.

For a future startup-capable `get_vehicle_last_trip()`, I would prefer:

```text
distance=0.1-
pageSize=1
→ follow _links.last
→ latest trip
```

rather than first fetching the global count and then calculating `indexRange=total-1-total-1`.

So my suggested scope remains:

- keep PR #616 focused on the generic raw historical transport;
- keep the optional `since` parameter;
- leave the current engine-stop `get_vehicle_last_trip()` unchanged in this PR;
- if upstream wants `get_vehicle_last_trip()` to also work as a startup/initialization helper, fix that separately (or fold it in only if the maintainer prefers).

I have also documented the measured pagination/index behavior and the relevant official Stellantis API links in the SV Dashboard repository so we have a durable reference for future API audits:

https://github.com/CaneTLOTW/sv_dashboard/blob/develop/docs/STELLANTIS_API_REFERENCE.md

CaneTLOTW:
@andreadegiovine 
Do you think this can be merged into main soon?

PR #619, opened 2026-09-17
Serialize MQTT client connect/disconnect and guard against a missing client

MoellerDi:
## Summary

`connect_mqtt()` and `async_shutdown()` both replace or tear down `self._mqtt` without any coordination between them, and `send_mqtt_message()` reads `self._mqtt` without checking it was even set.

- `connect_mqtt()` reuses the same `self._mqtt` object across reconnects (only building one when `self._mqtt is None`) and never calls `loop_stop()`. Since that check and the object creation both happen before the method's first `await`, several vehicle coordinators calling `connect_mqtt()` around the same time (`get_vehicle_status()` triggers one per coordinator) always end up on the same client object, calling `username_pw_set()` / `connect()` / `loop_start()` on it concurrently, which paho does not support:
  ```
  Coordinator A: self._mqtt is None                     -> creates client, assigns self._mqtt
  Coordinator A: await self._mqtt.connect(...)          # first await, suspends here
  Coordinator B: self._mqtt is not None (A's client)    -> reuses it, no new object
  Coordinator B: await self._mqtt.connect(...)          # same object, concurrently with A
  Coordinator A: self._mqtt.loop_start()
  Coordinator B: self._mqtt.loop_start()                # paho: "Loop already started"
  ```
  There is also no coordination with `async_shutdown()`'s own teardown of `self._mqtt`.
- `send_mqtt_message()` calls `self._mqtt.publish(...)` directly, with no check that `self._mqtt` is not `None`. Before the first `connect_mqtt()` ever succeeds, or after a dropped connection, this raises a plain `AttributeError` logged as "Unexpected error during MQTT message sending" instead of the actual, expected condition (MQTT is not connected).

This branch carries two commits:

1. `Serialize MQTT client connect and disconnect behind a lock`: adds `self._mqtt_lock` (`asyncio.Lock`) and holds it for the entire teardown+reconnect sequence, so `connect_mqtt()` / `reconnect_mqtt()` and `async_shutdown()` never operate on a client the other just replaced or tore down. `async_shutdown()` already tore its client down this way (drop `on_disconnect`, `disconnect()`, `loop_stop()` in the executor); that logic moves into a `_disconnect_mqtt_locked()` helper used by both call sites, including the connect path, which previously skipped it entirely. The `if self._mqtt is None:` / `if self._mqtt.is_connected():` checks that used to gate creation and disconnect are replaced by the always-safe helper. If the client is already connected once the lock is held, the call returns immediately instead of rebuilding it: `is_connected()` can briefly read `False` right after `loop_start()`, before the CONNACK is processed, so without this check a coordinator could tear down a connection that never had a chance to settle. `_on_mqtt_subscribe()`'s scheduled reconnect needs the opposite (rebuild even though the transport still looks connected), so this becomes two public methods, `connect_mqtt()` and `reconnect_mqtt()`, both thin wrappers around a private `_connect_mqtt(force: bool)`.
2. `Guard send_mqtt_message against a disconnected MQTT client`: adds the same "connect if needed" check `get_vehicle_status()` already does, plus a follow-up check after the connect attempt (since `connect_mqtt()` can fail or return early while shutting down). If the client is still not connected, it raises `CommunicationError` (this repo's existing exception for backend/transport failures) instead of letting an unrelated `AttributeError` surface.

## Change

`stellantis.py`, `StellantisVehicles`:

- `__init__`: adds `self._mqtt_lock = asyncio.Lock()` next to `self._mqtt`.
- `_disconnect_mqtt_locked()` (new): swaps `self._mqtt` out for `None`, drops its `on_disconnect` callback, disconnects it and joins its network thread (`loop_stop()`) in the executor. Caller must already hold `self._mqtt_lock`.
- `async_shutdown()`: the MQTT teardown block now runs inside `async with self._mqtt_lock:` and calls `_disconnect_mqtt_locked()`.
- `connect_mqtt()` / `reconnect_mqtt()` (public, thin wrappers) / `_connect_mqtt(force: bool)` (the actual implementation): the whole body of `_connect_mqtt` runs inside `async with self._mqtt_lock:`. The `_shutting_down` check moved inside the lock (a task can start before shutdown begins but only get the lock after `async_shutdown()` has already set the flag). Right after that, if `force` is not set and the client is already connected, the method returns `True` without touching it. Otherwise, teardown of any existing client goes through `_disconnect_mqtt_locked()`, then a fresh client is always created, configured and connected.
- `_on_mqtt_subscribe()`: its scheduled reconnect now calls `self.reconnect_mqtt()`.
- `send_mqtt_message()`: before building and publishing the message, connects if `self._mqtt` is `None` or not connected, then raises `CommunicationError` if it is still not usable afterwards. A dedicated `except CommunicationError:` clause logs this as a plain warning (an expected condition) instead of the generic `except Exception:` branch's full exception log.

## Test plan

Tested and verified manually against my own Home Assistant instance:

- [x] With a single vehicle configured, reload the integration and confirm the debug log shows a clean `connect_mqtt` / `_on_mqtt_connect` / `_on_mqtt_subscribe` sequence, and that MQTT-pushed updates actually arrive as `_on_mqtt_message` calls, not just polling.
- [x] With two or more vehicles configured, reload the integration and confirm the connection settles (no repeated connect/disconnect churn in the log right after startup) and MQTT push events keep arriving for all vehicles, not just the first one polled: this is the scenario the fix targets.
- [x] Send a remote command (e.g. lock/unlock) via a button/switch entity while MQTT is connected; confirm it succeeds and the command history updates.
- [x] Force an MQTT disconnect (e.g. briefly block the broker host) and confirm the integration reconnects on the next poll, with no "Loop already started" warnings in the log.
- [x] Reload or unload the integration (or restart Home Assistant) and confirm `async_shutdown()` cleanly tears down the MQTT client, no lingering paho thread warnings, and a subsequent reload reconnects normally.


MoellerDi:
All tests have been successfully completed. I’ll keep an eye on it for a bit longer before moving the PR out of draft status.

@andreadegiovine :
Do you have any comments or objections regarding this? Let me know if anything needs adjusting.

andreadegiovine:
Ok, when is ready mark it to be merged

MoellerDi:
@andreadegiovine this is ready for review now

PR #620, opened 2026-09-17
Reject a command while another is still in progress vehicle-side

### LINKED ISSUE #617: BUG ? : Obscure HA activity screenshots for Stellantis Buttons
### Mark

DS

### Model

DS N°4 Plug-in Hybrid 240

### Year

2025

### Engine

Hybrid

### Remote service

Connect Plus

### HomeAssistant version

 Core 2026.9.2 Supervisor 2026.09.2 Operating System 18.2 Interface utilisateur 20260729.7

### Integration version

2026.9.3 (2026.9.2-beta.2 at the moment when the issue was observed)

### What happened?

As activity dashboard displayed by HA was not clear at all (French translation at the minimum) I provide the activity downloaded file which clarify the effective activity and especially all buttons made unavailable following a start preconditionning action followed by a stop preconditionning action.

HA obscure activity screenshots (from most to less recent as the activity log):

Image
Image
Image
Image
Image
Image

HA activity downloaded file:

[2026-09-14 - Test préconditionnement - Journal activité.xlsx](https://github.com/user-attachments/files/32295835/2026-09-14.-.Test.preconditionnement.-.Journal.activite.xlsx)

MoellerDi:
~~Addresses the core cause behind andreadegiovine/homeassistant-stellantis-vehicles#617 (root-caused during triage, not a full fix of everything reported there, see Summary).~~

EDIT: ~~as confirmed by @chmtc94 this PR fixes #617~~

## Summary

`StellantisBaseEntity.available_command` (the base `available` for every action button: preconditioning, doors, lights, horn, wake-up, charge start/stop, charge limit on/off, …) used to fold in `StellantisVehicleCoordinator.pending_action`, which tracked whether a command was sent but had not yet received a final MQTT update:

```python
@property
def pending_action(self):
    """ Pending action. """
    if not self._commands_history:
        return False
    last_action_id = list(self._commands_history.keys())[-1]
    return not self._commands_history[last_action_id]["updates"]
```

This had two separate problems, both producing the same symptom: buttons appearing to have been "pressed" in the Home Assistant logbook when nobody touched them.

**1. `pending_action` looked at the single most recently sent command across the whole vehicle, not the one that mattered.** Pressing any one remote command made every command button unavailable until that one command's MQTT response came back, and when it did, every gated button flipped available again at the same instant, showing up in the logbook as several unrelated buttons all "pressed" together. Reported (with logbook screenshots and a HA activity export) in andreadegiovine/homeassistant-stellantis-vehicles#617.

**2. Hiding the pending state behind `available` produced spurious entries even for the one button actually pressed.** HA core's `ButtonEntity._async_press_action` already writes a real state change (a timestamp) the instant the button is pressed. Flipping `available` to `False` right after sending, and back to `True` once resolved, adds two more real state changes on top of that, each one rendered by the logbook as "was pressed" regardless of what actually happened.

## Change

One commit, in `base.py` (plus `const.py` for the new constants and the `de`/`en`/`fr`/`it` locale files for the new error message):

- `pending_action` becomes vehicle-wide again (no per-button scoping, as an earlier version of this PR tried), but is no longer wired into `available_command`. A button now blocks as soon as its command is sent (not waiting for any MQTT response), stays blocked through the intermediate lifecycle codes (`COMMAND_STATUS_STILL_IN_PROGRESS = ("900", "901", "903")`: accepted, vehicle asleep, forwarded to vehicle), and clears once any other code arrives (success or error) for that same command.
- A safety valve (`PENDING_ACTION_TIMEOUT = 60` seconds): if that many seconds pass since the blocking command's last recorded update (or since it was sent, if it never received one at all) without a further update, it stops blocking on its own, even though its actual final status never arrived.
- `send_command` raises a `ServiceValidationError` (new `command_already_pending` translation key, added to the `de`/`en`/`fr`/`it` locale files, naming both the command just attempted and the one still pending) instead of letting a second command through while one is already in flight; `available_command` is otherwise unchanged, so buttons stay visibly clickable throughout.

## Behaviour

- Pressing one remote command no longer disables unrelated buttons in the UI.
- A single button press now produces exactly one logbook entry, matching what actually happened.
- Sending a second command while one is still in flight vehicle-side (including while it is only "accepted"/"forwarded"/"vehicle asleep") is rejected locally with a clear, translated error instead of the button silently appearing unavailable.
- If a command's final status never arrives (e.g. a dropped MQTT connection), further commands are unblocked again once `PENDING_ACTION_TIMEOUT` seconds pass without any further update, rather than staying rejected forever.
- `_disabled_commands` (permanently-unsupported commands) and the MQTT-connectivity check in `available_command` are unchanged.

## Manual test plan

- Table-checked the `pending_action` logic against representative scenarios (blocking immediately on send, a second send being rejected while blocked, staying blocked through 900/901/903, clearing on a final code, and the timeout releasing the block after 60s with no update at all).
- [x] Press a button (e.g. `lights`) and confirm the logbook shows exactly one "was pressed" entry for it, not three.
- [x] Press a second, different button immediately after and confirm it is rejected with the translated "Cannot send ...: ... is still waiting for a response from the vehicle" error, naming the first button.
- [x] Let the first command reach its final status, then press a button again: confirm it is accepted normally.
- [x] Confirm a button pressed while offline/MQTT-disconnected still correctly shows unavailable (the `mqtt_is_connected` check in `available_command` is untouched).
- [x] Send a command, drop the MQTT connection before its final status arrives, and confirm further commands are rejected until `PENDING_ACTION_TIMEOUT` passes, then accepted again.


chmtc94:
@MoellerDi 

Great job (as usual)👍

I suggest two special cases to be checked:

1) Sending a second command different from the first one when the first one is still pending. Does the server (or the car) accept it?

2) Special case of previous case 1 when the second command is the opposite of the first one, that is a stop command when the corresponding start command is still pending.

PS: I still do not fully understand why the frontend logbook show buttons as "pressed" while the corresponding HA downloaded file via the same frontend list them as "unavailable" which are their real states 🤔 Your PR will be great for removing all these messy logs👍

EDIT : HACS button_card could be a good complement for Frontend handling of Stellantis Vehicle Commands 😉 👉 https://github.com/custom-cards/button-card

MoellerDi:
@chmtc94 

>Great job (as usual)👍

Thank you 😃 

> 1. Sending a second command different from the first one when the first one is still pending. Does the server (or the car) accept it?

The first version of this PR attempted this approach. However, it turned out that the API can only process one command at a time. Therefore, the integration allows only one command at a time.

> 2. Special case of previous case 1 when the second command is the opposite of the first one, that is a stop command when the corresponding start command is still pending.

The integration now blocks a new command while the first one is still being executed.

> PS: I still do not fully understand why the frontend logbook show buttons as "pressed" while the corresponding HA downloaded file via the same frontend list them as "unavailable" which are their real states 🤔 

I tried to explain it in the summary (second point).

MoellerDi:
>>PS: I still do not fully understand why the frontend logbook show buttons as "pressed" while the corresponding HA >>downloaded file via the same frontend list them as "unavailable" which are their real states 🤔
>
>I tried to explain it in the summary (second point).

A button in HA is a stateless entity. Consequently, if you check the "state" of a button, you simply get a timestamp. There is no state such as "unavailable" or "unknown"; there is only the timestamp of the last state change.

chmtc94:
> A button in HA is a stateless entity. Consequently, if you check the "state" of a button, you simply get a timestamp. There is no state such as "unavailable" or "unknown"; there is only the timestamp of the last state change.

I understand the concept of a stateless entity for HA buttons. However, I find it a bit (and more) confusing that HA displays "Pressed" in the HA activity dashboard, while showing "unavailable" in the activity log file downloaded from the HA activity dashboard menu ...

But anyway, your PR will completely eliminate these particularly confusing logs in the HA activity dashboard👍

chmtc94:
Confusing logbook should be fixed in HA now 😉
👉 https://github.com/home-assistant/frontend/pull/54203

chmtc94:
@MoellerDi 

Regardless of HA's plans to fix the confusing display in the activity log, avoiding the toggling of all other control buttons to "unavailable" and then "unknown" after a command finishes—which generates a flood of log events—struck me as a highly desirable additional change; it also masked the HA display issue by eliminating those events. 

I therefore regret a bit that this PR has been abandoned.

That said, it might still benefit from a specific visual treatment (such as a color change) for disabled buttons—one that is independent of the "unavailable" status.
That said a distinct treatment for a enabled/disabled status from available/unavailable HA status.

MoellerDi:
@chmtc94

Thanks for following up, and sorry for the confusion around closing it. I had assumed the frontend fix (#54203) would cover the whole problem, but you're right that it only fixes the mislabeling, not the underlying event flood (every other button flipping to `unavailable`/`unknown` when one command is accepted or finishes), and it obviously does nothing for the actual duplicate-command case at the vehicle/API level. So I've reopened it, for now as a draft while we figure out the last bit below.

On your suggestion for a distinct visual treatment for "blocked" vs. real `unavailable`: I'd rather not drive that off the button entities themselves, since that's exactly what caused the original mess (a button's `available` isn't meant to represent "another command is in flight", and touching it re-triggers the same flood of logbook entries this PR removes).

One idea I'm toying with: a small diagnostic entity, something like `binary_sensor.<vehicle>_command_pending`, that would mirror the same `pending_action` state the buttons now check internally. A `binary_sensor` has a real on/off state, so its own transitions would show up in the logbook as plain "turned on"/"turned off" rather than the confusing "was pressed" label, and it wouldn't touch the buttons at all, just give a dashboard something to react to. With a HACS card like [button-card](https://github.com/custom-cards/button-card) (which you mentioned earlier), you could in theory template a button's style off it to grey it out while a command is pending, without any of that logic living in the integration's core button behavior.

Not sure yet if that's worth adding or overkill for this PR. Does it sound like what you had in mind, or were you thinking of something more built-in / automatic rather than dashboard-side?

chmtc94:
@MoellerDi 
I need some time for analyzing and describing new obviously linked issue today where buttons no more came available again..

MoellerDi:
@chmtc94

Sounds like it could be the same root cause: the current `pending_action` only looks at the single most recently sent command, with no timeout at all, so if its response ever gets lost or misrouted for any reason, every button stays unavailable forever (nothing can unstick it, since sending a new command is blocked too).

Just in case it's relevant while you dig into it: this PR already includes a `PENDING_ACTION_TIMEOUT` (60s) as a safety net, so even if a command's response never arrives, blocking clears on its own after that window instead of getting stuck permanently. Happy to take a look once you've got more details on what you're seeing.

MoellerDi:
@chmtc94

Rebased onto current `develop` now, it's mergeable again. And per your analysis on #639: this PR's `PENDING_ACTION_TIMEOUT` (60s) is exactly what would have kept your buttons from staying `unavailable` indefinitely after that dropped `ThermalPrecond` publish, good real-world confirmation that it's worth having.

Also added the `binary_sensor.<vehicle>_command_pending` we talked about (see the addendum in the description above): mirrors `pending_action` for a dashboard to react to, without touching the buttons themselves.

chmtc94:
@MoellerDi 
That's exactly what I had in mind. Just need to find the best way to adapt the frontend accordingly and prevent flooding the logbook😉

